PR #3, opened 2016-02-02
consider all relevant commands in 'linked_dylibs'

UniqMartin:
Linked dylibs are not limited to the `LC_LOAD_DYLIB` load command. For completeness and consistency with `otool -L`, consider all relevant load commands (all with a `DylibCommand` structure, except `LC_ID_DYLIB`).

Also adjust `change_install_name` to work the same on the modification side and for consistency with `install_name_tool`.

One way to see a real-world issue due to this is to use `brew linkage qt5` (from [`homebrew/dev-tools`](https://github.com/Homebrew/homebrew-dev-tools)) on a normal `qt5` build. The output will be different depending on whether or not `HOMEBREW_RUBY_MACHO` is set because Qt uses weak linkage. (Discovered while testing Homebrew/homebrew#45001.)

(Not sure what your goals for testing are, so here's just the fix without any tests for now.)

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -55,6 +55,16 @@ module MachO
 		0x2e => :LC_LINKER_OPTIMIZATION_HINT
 	}
 
+	# load commands responsible for loading dylibs
+	# @api private
+	DYLIB_LOAD_COMMANDS = [
+		:LC_LOAD_DYLIB,
+		:LC_LOAD_WEAK_DYLIB,
+		:LC_REEXPORT_DYLIB,
+		:LC_LAZY_LOAD_DYLIB,
+		:LC_LOAD_UPWARD_DYLIB,
+	].freeze
+
 	# association of load command symbols to string representations of classes
 	# @api private
 	LC_STRUCTURES = {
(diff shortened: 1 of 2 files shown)


woodruffw:
Great!

Before I merge this, could you update the associations of `LC_LAZY_LOAD_DYLIB` and `LC_LOAD_UPWARD_DYLIB` in [`LC_STRUCTURES`](https://github.com/woodruffw/ruby-macho/blob/master/lib/macho/load_commands.rb#L60) to `DylibCommand`? I had set them as generic `LoadCommand`s because of scant documentation, but they need to be changed now that we're retrieving strings from them.

On testing, I'll experiment with some linker flags and see if I can whip up some binaries that have these load commands.

UniqMartin:
> Before I merge this, could you update the associations of `LC_LAZY_LOAD_DYLIB` and `LC_LOAD_UPWARD_DYLIB` in `LC_STRUCTURES` to `DylibCommand`?

Done. :smile_cat: Thanks for noticing! I assumed it was already like that, but hadn't checked/tested.

> On testing, I'll experiment with some linker flags and see if I can whip up some binaries that have these load commands.

The linker flags `-lazy_library` and `-upward_library` should help with that (see `man 1 ld` for the full list).

woodruffw:
> The linker flags -lazy_library and -upward_library should help with that (see man 1 ld for the full list).
Awesome, I'll make some binaries with these.

PR #5, opened 2016-02-03
eliminate CPU_TYPE_X86 in favor of CPU_TYPE_I386

UniqMartin:
`CPU_TYPE_X86` and `CPU_TYPE_I386` have the same numeric value, making the `CPU_TYPE_X86 => "CPU_TYPE_X86"` entry in `CPU_TYPES` useless as it is immediately overwritten in the following line.

Use `CPU_TYPE_I386` everywhere as that's the constant used in newer versions of the [`mach/machine.h` header](http://www.opensource.apple.com/source/cctools/cctools-877.5/include/mach/machine.h) (linked file is from `cctools-877.5`).

### DIFF
--- a/lib/macho/headers.rb
+++ b/lib/macho/headers.rb
@@ -33,14 +33,11 @@ module MachO
 	# any CPU (unused?)
 	CPU_TYPE_ANY = -1
 
-	# x86 compatible CPUs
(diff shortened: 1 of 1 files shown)


woodruffw:
Sounds good to me :+1:

PR #8, opened 2016-02-05
Always return unique list of dylibs and clarify comments

UniqMartin:
Here's an example of a non-fat binary that links multiple times to the same dylib:

- `$(brew --cellar)/go/1.5.3/libexec/bin/godoc`

Here are some examples of fat binaries provided by the system (on OS X 10.11.3) that link to different subsets of libraries depending on architecture (differences can be inspected with `otool -arch <arch> -L <file>`, where `<arch>` is one of `i386`, `x86_64`):

- `/System/Library/CoreServices/Photo Library Migration Utility.app/Contents/Frameworks/iLifeSlideshow.framework/Versions/A/iLifeSlideshow`
- `/System/Library/PrivateFrameworks/iTunesAccess.framework/Versions/A/iTunesAccess`
- `/System/Library/Video/Plug-Ins/AppleProResCodec.bundle/Contents/MacOS/AppleProResCodec`

### DIFF
--- a/lib/macho/fat_file.rb
+++ b/lib/macho/fat_file.rb
@@ -131,7 +131,9 @@ module MachO
 		# All shared libraries linked to the file's Mach-Os.
 		# @return [Array<String>] an array of all shared libraries
 		def linked_dylibs
-			# can machos inside fat binaries have different dylibs?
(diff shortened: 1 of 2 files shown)


woodruffw:
Looks great :+1:

Issue #7, opened 2016-02-04
Benchmark Ruby-macho vs otool? 


xu-cheng:
First of all, good work here. :bow:

Since now this is included in Homebrew, I'm curious to know the performance between Ruby implementation vs old `otool`. So how about adding benchmark along side test suite?

woodruffw:
Thank you!

Benchmarking is a great idea. Do you have any instrumentation in mind?

xu-cheng:
You may find https://github.com/evanphx/benchmark-ips useful.

woodruffw:
This looks great. I'm expanding the test suite on rarer/unusual load commands today, but I'll see about adding this tomorrow.

UniqMartin:
Certainly not a full-featured benchmark, but here are some numbers that give an indication of the status quo on crunching through a large number of files and identifying/dumping Mach-O files among them. (Spoiler: There's still room for improvement). Numbers obviously depend on a lot of factors, including what formulae are installed and disk speed. I'm showing numbers for three consecutive runs, so caching effects were taken into account to some degree.

Identifying Mach-O files:

```
$ unset HOMEBREW_RUBY_MACHO
$ brew time macho-find "$(brew --cellar)" > /dev/null
brew macho-find:   2.150s user,   1.670s system,   3.841s total ( 99.5% cpu)
brew macho-find:   2.240s user,   1.720s system,   3.974s total ( 99.7% cpu)
brew macho-find:   2.260s user,   1.770s system,   4.054s total ( 99.4% cpu)

$ export HOMEBREW_RUBY_MACHO=1
$ brew time macho-find "$(brew --cellar)" > /dev/null
brew macho-find:   6.790s user,   7.360s system,  30.213s total ( 46.8% cpu)
brew macho-find:   6.400s user,   7.000s system,  28.919s total ( 46.3% cpu)
brew macho-find:   6.560s user,   7.160s system,  28.965s total ( 47.4% cpu)
```

Dumping Mach-O files:

```
$ unset HOMEBREW_RUBY_MACHO
$ brew time macho-dump "$(brew --cellar)" > /dev/null
brew macho-dump:   9.470s user,  15.820s system,  40.027s total ( 63.2% cpu)
brew macho-dump:   8.120s user,  12.300s system,  21.513s total ( 94.9% cpu)
brew macho-dump:   8.390s user,  12.910s system,  22.518s total ( 94.6% cpu)

$ export HOMEBREW_RUBY_MACHO=1
[...]

PR #9, opened 2016-02-05
Testing: Add skeleton for new benchmarking tests.

woodruffw:
Add test_bench.rb, which will contain benchmarking tests
for both ruby-macho and otool/install_name_tool.

Add benchmark/ips as a dependency to conduct the forthcoming
benchmarks.

Add Helpers.installed? to selectively run benchmark tests
if external utilities are not available.

cc @UniqMartin, @xu-cheng 
ref #7 

- [x] Stub out benchmarks
- [x] Fill in `ruby-macho` benchmarks
- [x] Fill in `otool`/`install_name_tool` benchmarks

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,5 +3,5 @@ source 'https://rubygems.org'
 group :test do
 	gem 'rake'
 	gem 'minitest'
-	gem 'simplecov'
+	gem 'benchmark-ips'
 end
--- a/Rakefile
+++ b/Rakefile
@@ -1,4 +1,5 @@
 require "rake/testtask"
+require_relative "test/bench"
 
 Rake::TestTask.new do |t|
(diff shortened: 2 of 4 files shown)


woodruffw:
I'm going to continue to fill this in, just wanted some opinions on the structure.

woodruffw:
I've pushed some (very) preliminary benchmarks.

They can be run with `rake bench`.

On my machine (Mac Mini 2014), the results are very promising. Mach-O access operations (fetching IDs, install names) are generally 50 to 100 times faster with `ruby-macho` than with `otool`, and modification operations are generally 2-10 times faster than with `install_name_tool`. This is probably largely attributable to the significant overhead in frequently shelling out to the utilities. There's also still a great deal of room for optimization in `ruby-macho`.

PR #11, opened 2016-02-05
MagicError: don't fail if magic number is nil

UniqMartin:
If the read file is too short (or empty), `magic` will be nil and the `MagicError` exception itself will throw an exception because nil cannot be formatted as an integer. Provide a more appropriate exception message in those cases.

@woodruffw I tagged this as `in progress` because I'm not sure this is a good fix for the issue and would appreciate your feedback. Should we create a separate superclass for exceptions that don't indicate an error in parsing, but simply the fact that the file is neither a Mach-O nor a fat binary? Possible reasons might be:

- It doesn't have the expected magic number.
- It's too short to accommodate the headers, so it cannot be a Mach-O/fat binary.
- It does have the expected magic number, but other sanity checks fail. (Can happen if the magic number is shared by other file formats, too.)

That might simplify handling/suppressing these exceptions where appropriate while still getting useful errors if there are real problems.

### DIFF
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -5,9 +5,13 @@ module MachO
 
   # Raised when a file's magic bytes are not valid Mach-O magic.
   class MagicError < MachOError
-    # @param num [Fixnum] the unknown number
+    # @param num [Fixnum, nil] the unknown number or nil if no number was read
     def initialize(num)
-      super "Unrecognized Mach-O magic: 0x#{"%02x" % num}"
+      if num
+        super "Unrecognized Mach-O magic: 0x#{"%02x" % num}"
+      else
+        super "Unrecognized Mach-O magic: file too short"
+      end
     end
   end
 


woodruffw:
An exception for non-parser errors sounds good, how about `NotAMachOError` or `FileSanityError`?

UniqMartin:
> An exception for non-parser errors sounds good, how about `NotAMachOError` or `FileSanityError`?

I like the former. Do you think we should stick with one exception class and just customize the message or do we really create a hierarchy like `MagicError < NotAMachOError`, `NotAMachOError < MachOError`?

> Fat little-endian Mach-Os have the same magic as Java classfiles, so this is definitely a good idea to avoid unfortunate mishaps with `.class`es installed by Homebrew.

That's exactly the shared magic number I had in mind. :grinning: Glad you're aware of it! Here's a snippet from my (non-public) Mach-O parser and how I decided to handle it. What do you think about this approach?

```ruby
def mach_o_binary?(path, *accept_kind)
  peek = path.read(32)
  magic, = peek[0..3].unpack("N")
  case magic
  when FatHeader::FAT_MAGIC
    # 0xcafebabe is also used by Java class files. Fortunately, the four
    # bytes directly following the magic value encode the file format
    # version there, and the lowest possible version is 43. Having more
    # than 30 different architectures is quite unlikely, so that should be
    # a practical limit to distinguish between the two.
    nfat_arch, = peek[4..7].unpack("N")
    return false if nfat_arch < 1 || nfat_arch > 30

  # …
```

woodruffw:
> I like the former. Do you think we should stick with one exception class and just customize the message or do we really create a hierarchy like MagicError < NotAMachOError, NotAMachOError < MachOError?

Subclassing `MagicError` makes the most sense to me, even through we're technically looking past the magic bytes for the purpose of distinguishing `.class`es.


> That's exactly the shared magic number I had in mind. :grinning: Glad you're aware of it! Here's a snippet from my (non-public) Mach-O parser and how I decided to handle it. What do you think about this approach?

This looks great to me!
In a perfect world I'd prefer a way to distinguish from classfiles without assuming that Fat Mach-Os have less than 30 archs, but a cursory glance at the classfile spec suggests that that's unlikely to ever happen.

:+1:  all around

UniqMartin:
@woodruffw Sorry for the inactivity here; I'm a bit too busy lately! If you have more time to address this, feel free to do that. I don't mind abandoning this PR in favor of your solution. (But I'm happy to review if you post it as a PR first.)

Getting a version of this merged would be helpful to advance on Homebrew/homebrew#49091 (in combination with a modified Homebrew/homebrew#48817 that fails hard instead of just an `onoe`).

woodruffw:
> Sorry for the inactivity here; I'm a bit too busy lately!

Don't sweat it - I've been busy too.

I'll try to open a PR with more specific exceptions today or tomorrow (schoolwork permitting).

woodruffw:
Opened the changes as #18

UniqMartin:
> Opened the changes as #18

Thanks for taking care of this! :bow:

PR #12, opened 2016-02-05
FatFile/MachOFile: avoid some code duplication

UniqMartin:
The commits go into a bit more detail, but the idea is to rely more on `MachOStructure.new_from_bin` for handling the structs instead of replicating that logic (and basically ignoring `FORMAT`), thus making it more likely to create inconsistencies.

*For full disclosure where I hope this leads to:* This is one (of possibly several) preparatory PRs to pave the way for full endian awareness in the code. In a local branch I have already hacked together everything so that the code is able to correctly read 32/64-bit x86/PPC/ARM binaries. (I cross-checked with `otool` to a reasonable extent and let it crunch through most of my system disk.)

### DIFF
--- a/lib/macho/headers.rb
+++ b/lib/macho/headers.rb
@@ -235,7 +235,9 @@ module MachO
 				flags)
 			@magic = magic
 			@cputype = cputype
-			@cpusubtype = cpusubtype
+			# For now we're not interested in additional capability bits also to be
+			# found in the `cpusubtype` field. We only care about the CPU sub-type.
+			@cpusubtype = cpusubtype & ~CPU_SUBTYPE_MASK
 			@filetype = filetype
 			@ncmds = ncmds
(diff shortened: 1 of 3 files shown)


lib/macho/headers.rb
```diff
@@ -162,7 +162,7 @@ class FatHeader < MachOStructure
 		# @return [Fixnum] the number of fat architecture structures following the header
 		attr_reader :nfat_arch
 
-		FORMAT = "VV"
+		FORMAT = "N2" # always big-endian
```
woodruffw: Good catch here, this was a big error on my part.

woodruffw:
Looks great to me! Merge at your leisure.

UniqMartin:
> Looks great to me! Merge at your leisure.

Thanks, just did! I rebased/pushed it beforehand to test the new CI goodness. :smile_cat:

PR #14, opened 2016-02-07
1.8.7 Compatibility

woodruffw:
Tasks:

- [x] Get the test suite itself compatible with 1.8.7.
- [x] Make `ruby-macho` itself compatible with 1.8.7.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,7 @@
+$:.unshift "#{File.dirname(__FILE__)}/lib"
+$:.unshift "#{File.dirname(__FILE__)}/test"
 require "rake/testtask"
-require_relative "test/bench"
+require "bench"
 
 Rake::TestTask.new do |t|
 	t.libs << 'test'
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,5 @@ rvm:
     - 1.8.7
     - 1.9.3
     - 2.0.0
-matrix:
-    allow_failures:
-        - rvm: 1.8.7
 cache: bundler
 sudo: false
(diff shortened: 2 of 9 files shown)


woodruffw:
cc @UniqMartin:

Any objections to this partially reverting #2? I think we discussed adding an *extend* hierarchy, but since `flat_map` is currently the only incompatibility I think it makes sense to change it for now.

xu-cheng:
FYI, `flat_map` is back ported inside Homebrew.

woodruffw:
> FYI, flat_map is back ported inside Homebrew.

Yeah, @UniqMartin mentioned that. I don't want to assume that `ruby-macho` is only being run in the context of Homebrew (being available as a gem and all), so it's going to need its own collection of backports eventually. When that day comes, I'll probably take them directly from Homebrew.

Rakefile
```diff
@@ -1,5 +1,7 @@
+$:.unshift "lib"
+$:.unshift "test"
```
UniqMartin: You might want to use absolute paths (possibly derived from the location of the `Rakefile`) instead. Depending on when things are `require`d in the code, otherwise an innocent looking `cd` might unexpectedly break things.

UniqMartin:
One comment, otherwise :+1:.

> Any objections to this partially reverting #2?

Don't worry about that. This partial revert is totally warranted, if it even deserves to be called “revert”. After all, it's only substituting `flat_map` with `map` and `flatten`. (Plus I introduced this incompatibility and then was impertinent enough to ask for 1.8.7 compatibility. :smirk_cat:)

Rakefile
woodruffw: Made it absolute, thanks!

woodruffw:
Awesome! I'll merge this after the CI completes :tada:

PR #16, opened 2016-02-14
Maintenance: Replace tab indentation with two-space.

woodruffw:
Everybody else in the Ruby world uses two-space, so we
should too.


UniqMartin:
There are still tabs in `Gemfile`, `README.md`, `Rakefile`, `ruby-macho.gemspec`, and `test/src/{,lib}hello.c`. Is that intentional or an oversight?

woodruffw:
Oops, oversight. I'll fix that right now.

PR #18, opened 2016-02-16
Error Handling: Add NotAMachOError for edge cases.

woodruffw:
There are two edge cases early-on in the parsing process:
files too small to have a valid magic number and files
with valid magic but invalid contents (most notably, Java
classfiles).

The NotAMachOError class covers both of these cases, and is raised
in FatFile#get_fat_header and MachOFile#get_and_check_magic.

cc @UniqMartin

### DIFF
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -3,11 +3,33 @@ module MachO
   class MachOError < RuntimeError
   end
 
+  # Raised when a file is not a Mach-O.
+  class NotAMachOError < MachOError
+    # @param error [String] the error in question
+    def initialize(error)
+      super error
+    end
+  end
+
+  # Raised when a file is too short to be a valid Mach-O file.
+  class TruncatedFileError < NotAMachOError
+    def initialize
+      super "File is too short to be a valid Mach-O"
+    end
+  end
+
   # Raised when a file's magic bytes are not valid Mach-O magic.
-  class MagicError < MachOError
+  class MagicError < NotAMachOError
     # @param num [Fixnum] the unknown number
     def initialize(num)
       super "Unrecognized Mach-O magic: 0x#{"%02x" % num}"
+    end
+  end
+
+  # Raised when a file is a Java classfile instead of a fat Mach-O.
+  class JavaClassFileError < NotAMachOError
+    def initialize
+      super "File is a Java class file"
     end
   end
 
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -314,8 +314,12 @@ module MachO
     # The file's Mach-O header structure.
     # @return [MachO::MachHeader] if the Mach-O is 32-bit
     # @return [MachO::MachHeader64] if the Mach-O is 64-bit
+    # @raise [MachO::TruncatedFileError] if the file is too small to have a valid header
     # @private
     def get_mach_header
+      # the smallest Mach-O header is 28 bytes
+      raise TruncatedFileError.new if @raw_data.size < 28
+
       magic = get_and_check_magic
       mh_klass = MachO.magic32?(magic) ? MachHeader : MachHeader64
       mh = mh_klass.new_from_bin(@raw_data[0, mh_klass.bytesize])
--- a/test/helpers.rb
+++ b/test/helpers.rb
@@ -1,6 +1,7 @@
 require "macho"
 require "digest/sha1"
 require "fileutils"
+require "tempfile"
 
 module Helpers
   TEST_OBJ = "test/bin/hello.o"
@@ -43,5 +44,13 @@ module Helpers
     checks.delete(except)
 
     checks
+  end
+
+  def tempfile_with_data(filename, data)
+    Tempfile.open(filename) do |file|
+      file.write(data)
+      file.rewind
+      yield file
+    end
   end
 end
--- a/test/test_fat.rb
+++ b/test/test_fat.rb
@@ -5,6 +5,31 @@ require "macho"
 class FatFileTest < Minitest::Test
   include Helpers
 
+  def test_empty_file
+    tempfile_with_data("empty_file", "") do |empty_file|
+      assert_raises MachO::TruncatedFileError do
+        MachO::FatFile.new(empty_file.path)
+      end
+    end
+  end
+
+  def test_truncated_file
+    tempfile_with_data("truncated_file", "\xCA\xFE\xBA\xBE\x00\x00") do |truncated_file|
+      assert_raises MachO::TruncatedFileError do
+        MachO::FatFile.new(truncated_file.path)
+      end
+    end
+  end
+
+  def test_java_classfile
+    blob = "\xCA\xFE\xBA\xBE\x00\x00\x00\x33therestofthisfileisnotarealjavaclassfile"
+    tempfile_with_data("fake_java_class_file", blob) do |fake_java_class_file|
+      assert_raises MachO::JavaClassFileError do
+        MachO::FatFile.new(fake_java_class_file.path)
+      end
+    end
+  end
+
   def test_fat_header
     file = MachO::FatFile.new(TEST_FAT_EXE)
     header = file.header
--- a/test/test_macho.rb
+++ b/test/test_macho.rb
@@ -5,6 +5,22 @@ require "macho"
 class MachOFileTest < Minitest::Test
   include Helpers
 
+  def test_empty_file
+    tempfile_with_data("empty_file", "") do |empty_file|
+      assert_raises MachO::TruncatedFileError do
+        MachO::MachOFile.new(empty_file.path)
+      end
+    end
+  end
+
+  def test_truncated_file
+    tempfile_with_data("truncated_file", "\xFE\xED\xFA\xCE\x00\x00") do |truncated_file|
+      assert_raises MachO::TruncatedFileError do
+        MachO::MachOFile.new(truncated_file.path)
+      end
+    end
+  end
+
   def test_load_commands
     file = MachO::MachOFile.new(TEST_EXE)
 
--- a/lib/macho/fat_file.rb
+++ b/lib/macho/fat_file.rb
@@ -179,14 +179,28 @@ module MachO
 
     # Obtain the fat header from raw file data.
     # @return [MachO::FatHeader] the fat header
+    # @raise [MachO::TruncatedFileError] if the file is too small to have a valid header
     # @raise [MachO::MagicError] if the magic is not valid Mach-O magic
     # @raise [MachO::MachOBinaryError] if the magic is for a non-fat Mach-O file
+    # @raise [MachO::JavaClassFileError] if the file is a Java classfile
     # @private
     def get_fat_header
+      # the smallest fat Mach-O header is 8 bytes
+      raise TruncatedFileError.new if @raw_data.size < 8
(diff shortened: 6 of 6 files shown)


lib/macho/macho_file.rb
```diff
@@ -329,10 +329,13 @@ def get_mach_header
 
     # Read just the file's magic number and check its validity.
     # @return [Fixnum] the magic
+    # @raise [MachO::NotAMachOError] if the file is too small to have valid magic
     # @raise [MachO::MagicError] if the magic is not valid Mach-O magic
     # @raise [MachO::FatBinaryError] if the magic is for a Fat file
     # @private
     def get_and_check_magic
+      raise NotAMachOError.new("file is too small to be a Mach-O") if @raw_data.size < 4
```
UniqMartin: I think the limit should be 28 bytes (the size of `MachHeader`) instead of 4. I'd like to see a similar check in `FatFile#get_fat_header` with a limit of 8 bytes (the size of `FatHeader`; or 8+20+28 bytes if we consider a fat binary with no embedded Mach-O binaries to be invalid). See also my (yet to be written) non-code comment for more details.

test/test_macho.rb
```diff
@@ -5,6 +5,12 @@
 class MachOFileTest < Minitest::Test
   include Helpers
 
+  def test_empty_file
+    assert_raises MachO::NotAMachOError do
+      MachO::MachOFile.new_from_bin("\x00")
```
UniqMartin: I think this should be just `""` (an empty string) to stay true to the name of the test.

lib/macho/macho_file.rb
woodruffw: This makes sense.

As a side note, do you know if `clang`/any part of the OS X toolchain actually ever produces just the Mach-O header (with `ncmds` and `sizeofcmds` = 0) and nothing else?

test/test_macho.rb
woodruffw: Good point, changing it now :+1:

UniqMartin:
**Exception Classes.** Can we make it such that there is a single exception class that can be caught for cases where the file was successfully read, but turned out not to be a Mach-O or fat binary? I can think of three possible hierarchies that make sense to me (ordered by personal preference):

```
-+ NotAMachOError        # common ancestor, but not actually raised
 |-- MagicError          # used as before
 |-- TruncatedFileError  # too short
 |-- JavaClassFileError  # Java class file
 \-- …                   # more exceptions for cases we haven't thought of yet

-+ NotAMachOError        # too short, Java class file, …
 \-- MagicError          # used as before

-+ NotAMachOError        # common ancestor, but not actually raised
 |-- MagicError          # used as before
 \-- NonMagicError       # too short, Java class file, …
```

The important thing is that it would be sufficient to rescue `NotAMachOError` and interpret it as “regular file” (the use case in Homebrew's extension of `Pathname`), no matter what the exact cause is and without ignoring actual issues (e.g. due to unknown or misinterpreted load commands).

----

**Additional Corner Cases.** Can you add the following test cases and adjust the library code such that they succeed? (You might need to adjust the names of the exception classes per the first part of this comment.) In `test/test_fat.rb`:

```ruby
def test_empty_file
  assert_raises MachO::NotAMachOError do
    MachO::FatFile.new_from_bin("")
  end
end

def test_truncated_file
  assert_raises MachO::NotAMachOError do
    MachO::FatFile.new_from_bin("\xCA\xFE\xBA\xBE\x00\x00")
  end
end
```

In `test/test_macho.rb`:

```ruby
def test_truncated_file
  assert_raises MachO::NotAMachOError do
    MachO::MachOFile.new_from_bin("\xFE\xED\xFA\xCE\x00\x00")
  end
end
```

test/test_macho.rb
UniqMartin: :cool:

woodruffw:
The first hierarchy looks good to me!

lib/macho/macho_file.rb
UniqMartin: I'm not aware of a way to create such a file except of hand-crafting it.

woodruffw:
> MachO::FatFile.new_from_bin("\xCA\xFE\xBA\xBE\x00\x00")

`FatFile` doesn't actually have a `new_from_bin` method, so I'm going to add another truncated classfile to *test/bin* (and add the missing method in a later commit).

lib/macho/exceptions.rb
```diff
@@ -3,14 +3,36 @@ module MachO
...
+    # @param error [String] the error in question
+    def initialize(error)
+      super error
+    end
+  end
+
+  # Raised when a file is too short to be a valid Mach-O file.
+  class TruncatedFileError < NotAMachOError
+    def initialize
+      "File is too short to be a valid Mach-O"
```
UniqMartin: The `super` is missing here.

lib/macho/exceptions.rb
```diff
@@ -3,14 +3,36 @@ module MachO
...
     # @param num [Fixnum] the unknown number
     def initialize(num)
       super "Unrecognized Mach-O magic: 0x#{"%02x" % num}"
     end
   end
 
+  # Raised when a file is a Java classfile instead of a fat Mach-O.
+  class JavaClassFileError < NotAMachOError
+    def initialize
+      "File is a Java class file"
```
UniqMartin: The `super` is missing here, too.

lib/macho/exceptions.rb
woodruffw: Fixed, both this and below.

test/test_fat.rb
```diff
@@ -5,6 +5,40 @@
 class FatFileTest < Minitest::Test
   include Helpers
 
+  def test_empty_file
+    empty_file = tempfile_with_data("empty_file", "")
+
+    assert_raises MachO::NotAMachOError do
```
UniqMartin: Maybe check for the more specific exception classes like `TruncatedFileError` here and in the other newly added test cases?

test/test_fat.rb
```diff
@@ -5,6 +5,40 @@
...
+    truncated_file.unlink
+  end
+
+  def test_java_classfile
+    fake_java_class_file = tempfile_with_data("fake_java_class_file", "\xCA\xFE\xBA\xBE\x00\x00\x00\x33therestofthisfileisnotarealjavaclassfile")
+
+    assert_raises MachO::NotAMachOError do
+      MachO::FatFile.new(fake_java_class_file)
+    end
+
```
UniqMartin: This empty line is odd and not in line with the other methods.

test/test_macho.rb
```diff
@@ -5,6 +5,29 @@
...
+    empty_file.unlink
+  end
+
+  def test_truncated_file
+    truncated_file = tempfile_with_data("truncated_file", "\xFE\xED\xFA\xCE\x00\x00")
+
+    assert_raises MachO::NotAMachOError do
+      MachO::MachOFile.new(truncated_file)
+    end
+
```
UniqMartin: This empty line is odd and not in line with the other methods.

test/test_fat.rb
```diff
@@ -5,6 +5,40 @@
...
 
+  def test_empty_file
+    empty_file = tempfile_with_data("empty_file", "")
+
+    assert_raises MachO::NotAMachOError do
+      MachO::FatFile.new(empty_file)
+    end
+  ensure
+    empty_file.close
+    empty_file.unlink
```
UniqMartin: I think it would be good to get rid of this code duplication in all the `ensure`s. An untested idea would be to modify `tempfile_with_data` to remember the generated file (in an instance variable) and then clean it up in the `teardown` method (not sure how to put that into `Helpers` correctly so that test classes can still add their own `teardown` methods).

UniqMartin:
> `FatFile` doesn't actually have a `new_from_bin` method, so I'm going to add another truncated classfile to *test/bin* (and add the missing method in a later commit).

Totally missed that `FatFile` doesn't have that method, sorry. I think adding it (in a later commit) is still a good idea as it will make the interface for both classes more similar.

> Ended up creating a `Helpers.tempfile_with_data` to handle these tests.

That's also a good solution and probably more broadly applicable.

test/test_fat.rb
woodruffw: I agree. I could probably turn `tempfile_with_data` into a block, that would reduce the repetition and avoid confusion with the interaction between `Helpers` and each test class.

UniqMartin: Yes, turning `tempfile_with_data` into a method that accepts a block that performs the actual test that involves the temporary file also sounds like a good solution.

woodruffw:
The new `Helpers.tempfile_with_data` takes a block that handles setup and teardown for each temp file, so that should less of a hassle to work with.

test/test_fat.rb
```diff
@@ -5,6 +5,30 @@
...
+  def test_truncated_file
+    tempfile_with_data("truncated_file", "\xCA\xFE\xBA\xBE\x00\x00") do |truncated_file|
+      assert_raises MachO::NotAMachOError do
+        MachO::FatFile.new(truncated_file)
+      end
+    end
+  end
+
+  def test_java_classfile
+    tempfile_with_data("fake_java_class_file", "\xCA\xFE\xBA\xBE\x00\x00\x00\x33therestofthisfileisnotarealjavaclassfile") do |fake_java_class_file|
```
UniqMartin: Maybe split this into two lines to avoid an awfully long line?

```ruby
blob = "\xCA\xFE\xBA\xBE\x00\x00\x00\x33therestofthisfileisnotarealjavaclassfile"
tempfile_with_data("fake_java_class_file", blob) do |fake_java_class_file|
```

woodruffw: Yeah, fixing it now.

test/helpers.rb
```diff
@@ -44,4 +45,17 @@ def filechecks(except = nil)
...
+    file.write(data)
+    file.rewind
+
+    begin
+      yield file
+    ensure
+      file.close
+      file.unlink
+    end
+  end
```
UniqMartin: I think you can simplify this further to (the only difference being that the file won't be immediately deleted, see the `Tempfile` documentation):

```ruby
def tempfile_with_data(filename, data)
  Tempfile.open(filename) do |file|
    file.write(data)
    file.rewind
    yield file
  end
end
```

woodruffw: That's much cleaner, thanks!

UniqMartin:
There is one (now hidden) code comment about the exception classes in the test suite, that you haven't addressed or replied to, but that's not at all a blocker, just a detail:

> Maybe check for the more specific exception classes like `TruncatedFileError` (instead of their common ancestor) here and in the other newly added test cases?

Overall I'm quite happy with how things look by now. Feel free to merge if you and the CI are happy. :+1:

woodruffw:
> Maybe check for the more specific exception classes like TruncatedFileError (instead of their common ancestor) here and in the other newly added test cases?

Oops, missed that. That's an important detail.

After that (and the CI passes), I'll merge :smile:

UniqMartin:
> That's an important detail.

I'm glad that we agree on this. :smile_cat:

> After that (and the CI passes), I'll merge :smile:

:+1: :clap:

woodruffw:
(Afternote: I added `new_from_bin` in 9cdca3f7a012e0b7fded8b018bc3c9c073a7a250)

PR #21, opened 2016-02-18
Testing: Simplify running the test suite

UniqMartin:
While testing locally with a bunch of different Rubies I found it quite cumbersome to set everything up. The result is `run-tests` which for me simplifies things down to a `with-ruby 1.8 ./run-tests`. As a kind of precondition I felt compelled to re-organize `.gitignore` and sort out the `Gemfile.lock` situation. (I hope I haven't missed anything by moving around and anchoring the ignore rules.)

@woodruffw Not sure what your local setup looks like, so I'm not sure if I'm only doing myself a favor and cluttering up your repository root or if my script is useful enough to occupy that spot.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,16 @@
-doc/
-.yardoc/
-coverage/
-*.dylib
-*.so
-*.a
-*.bin
-*.gem
-!test/bin/*
-Gemfile.lock
+# Documentation.
+/.yardoc/
+/doc/
+
+# Gems for redistribution.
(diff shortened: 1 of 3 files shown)


woodruffw:
This looks good to me! My local setup is just `rvm` + `bundle exec`, so this is a welcome change.

UniqMartin:
Great! Thanks for merging! :smile_cat:

woodruffw:
No problem :+1:

PR #24, opened 2016-06-01
PPC (general big-endian) support

woodruffw:
- [x] Big-endian support implemented
- [x] PPC constants added
- [x] PPC single-arch tests
- [x] PPC fat tests

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -73,7 +73,7 @@ module MachO
     :LC_SEGMENT => "SegmentCommand",
     :LC_SYMTAB => "SymtabCommand",
     :LC_SYMSEG => "LoadCommand", # obsolete
-    :LC_THREAD => "ThreadCommand",
+    :LC_THREAD => "ThreadCommand", # seems obsolete, but not documented as such
     :LC_UNIXTHREAD => "ThreadCommand",
     :LC_LOADFVMLIB => "LoadCommand", # obsolete
     :LC_IDFVMLIB => "LoadCommand", # obsolete
@@ -157,18 +157,23 @@ module MachO
     # @return [Fixnum] the size of the load command, in bytes
     attr_reader :cmdsize
 
-    FORMAT = "VV"
+    FORMAT = "L=2"
     SIZEOF = 8
 
     # Creates a new LoadCommand given an offset and binary string
+    # @param raw_data [String] the raw Mach-O data
+    # @param endianness [Symbol] the endianness of the command (:big or :little)
     # @param offset [Fixnum] the offset to initialize with
     # @param bin [String] the binary string to initialize with
     # @return [MachO::LoadCommand] the new load command
     # @api private
-    def self.new_from_bin(raw_data, offset, bin)
-      self.new(raw_data, offset, *bin.unpack(self::FORMAT))
+    def self.new_from_bin(raw_data, endianness, offset, bin)
+      format = specialize_format(self::FORMAT, endianness)
+
+      self.new(raw_data, offset, *bin.unpack(format))
     end
 
+    # @param raw_data [String] the raw Mach-O data
     # @param offset [Fixnum] the offset to initialize with
     # @param cmd [Fixnum] the load command's identifying number
     # @param cmdsize [Fixnum] the size of the load command in bytes
@@ -226,7 +231,7 @@ module MachO
     # @return [Array<Fixnum>] the UUID
     attr_reader :uuid
 
-    FORMAT = "VVa16"
+    FORMAT = "L=2a16"
     SIZEOF = 24
 
     # @api private
@@ -277,7 +282,7 @@ module MachO
     # @return [Fixnum] any flags associated with the segment
     attr_reader :flags
 
-    FORMAT = "VVa16VVVVVVVV"
+    FORMAT = "L=2a16L=4l=2L=2"
     SIZEOF = 56
 
     # @api private
@@ -336,7 +341,7 @@ module MachO
     # @return [Fixnum] any flags associated with the segment
     attr_reader :flags
 
-    FORMAT = "VVa16QQQQVVVV"
+    FORMAT = "L=2a16Q=4l=2L=2"
     SIZEOF = 72
 
     # @api private
@@ -381,7 +386,7 @@ module MachO
     # @return [Fixnum] the library's compatibility version number
     attr_reader :compatibility_version
 
-    FORMAT = "VVVVVV"
+    FORMAT = "L=6"
     SIZEOF = 24
 
     # @api private
@@ -402,7 +407,7 @@ module MachO
     # @return [MachO::LoadCommand::LCStr] the dynamic linker's path name as an LCStr
     attr_reader :name
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -424,7 +429,7 @@ module MachO
     # @return [Fixnum] a bit vector of linked modules
     attr_reader :linked_modules
 
-    FORMAT = "VVVVV"
+    FORMAT = "L=5"
     SIZEOF = 20
 
     # @api private
@@ -439,7 +444,8 @@ module MachO
   # A load command used to represent threads.
   # @note cctools-870 has all fields of thread_command commented out except common ones (cmd, cmdsize)
   class ThreadCommand < LoadCommand
-
+    FORMAT = "L=2"
+    SIZEOF = 8
   end
 
   # A load command containing the address of the dynamic shared library
@@ -470,7 +476,7 @@ module MachO
     # @return [void]
     attr_reader :reserved6
 
-    FORMAT = "VVVVVVVVVV"
+    FORMAT = "L=10"
     SIZEOF = 40
 
     # @api private
@@ -517,7 +523,7 @@ module MachO
     # @return [void]
     attr_reader :reserved6
 
-    FORMAT = "VVQQQQQQQQ"
+    FORMAT = "L=2Q=8"
     SIZEOF = 72
 
     # @api private
@@ -542,7 +548,7 @@ module MachO
     # @return [MachO::LoadCommand::LCStr] the umbrella framework name as an LCStr
     attr_reader :umbrella
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -558,7 +564,7 @@ module MachO
     # @return [MachO::LoadCommand::LCStr] the subumbrella framework name as an LCStr
     attr_reader :sub_umbrella
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -574,7 +580,7 @@ module MachO
     # @return [MachO::LoadCommand::LCStr] the sublibrary name as an LCStr
     attr_reader :sub_library
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -590,7 +596,7 @@ module MachO
     # @return [MachO::LoadCommand::LCStr] the subclient name as an LCStr
     attr_reader :sub_client
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -615,7 +621,7 @@ module MachO
     # @return the string table size in bytes
     attr_reader :strsize
 
-    FORMAT = "VVVVVV"
+    FORMAT = "L=6"
     SIZEOF = 24
 
     # @api private
@@ -686,7 +692,7 @@ module MachO
     attr_reader :nlocrel
 
 
-    FORMAT = "VVVVVVVVVVVVVVVVVVVV"
+    FORMAT = "L=20"
     SIZEOF = 80
 
     # ugh
@@ -726,7 +732,7 @@ module MachO
     # @return [Fixnum] the number of hints in the hint table
     attr_reader :nhints
 
-    FORMAT = "VVVV"
+    FORMAT = "L=4"
     SIZEOF = 16
 
     # @api private
@@ -743,7 +749,7 @@ module MachO
     # @return [Fixnum] the checksum or 0
     attr_reader :cksum
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -760,7 +766,7 @@ module MachO
     # @return [MachO::LoadCommand::LCStr] the path to add to the run path as an LCStr
     attr_reader :path
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -780,7 +786,7 @@ module MachO
     # @return [Fixnum] size of the data in the __LINKEDIT segment
     attr_reader :datasize
 
-    FORMAT = "VVVV"
+    FORMAT = "L=4"
     SIZEOF = 16
 
     # @api private
@@ -803,7 +809,7 @@ module MachO
     # @return [Fixnum] the encryption system, or 0 if not encrypted yet
     attr_reader :cryptid
 
-    FORMAT = "VVVVV"
+    FORMAT = "L=5"
     SIZEOF = 20
 
     # @api private
@@ -830,7 +836,7 @@ module MachO
     # @return [Fixnum] 64-bit padding value
     attr_reader :pad
 
-    FORMAT = "VVVVVV"
+    FORMAT = "L=6"
     SIZEOF = 24
 
     # @api private
@@ -852,7 +858,7 @@ module MachO
     # @return [Fixnum] the SDK version X.Y.Z packed as x16.y8.z8
     attr_reader :sdk
 
-    FORMAT = "VVVV"
+    FORMAT = "L=4"
     SIZEOF = 16
 
     # @api private
@@ -919,7 +925,7 @@ module MachO
     # @return [Fixnum] the size of the export information
     attr_reader :export_size
 
-    FORMAT = "VVVVVVVVVVVV"
+    FORMAT = "L=12"
     SIZEOF = 48
 
     # @api private
@@ -946,7 +952,7 @@ module MachO
     # @return [Fixnum] the number of strings
     attr_reader :count
 
-    FORMAT = "VVV"
+    FORMAT = "L=3"
     SIZEOF = 12
 
     # @api private
@@ -964,7 +970,7 @@ module MachO
     # @return [Fixnum] if not 0, the initial stack size.
     attr_reader :stacksize
 
-    FORMAT = "VVQQ"
+    FORMAT = "L=2Q=2"
     SIZEOF = 24
 
     # @api private
@@ -981,7 +987,7 @@ module MachO
     # @return [Fixnum] the version packed as a24.b10.c10.d10.e10
     attr_reader :version
 
-    FORMAT = "VVQ"
+    FORMAT = "L=2Q=1"
     SIZEOF = 16
 
     # @api private
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -8,6 +8,9 @@ module MachO
     # @return [String] the filename loaded from, or nil if loaded from a binary string
     attr_accessor :filename
 
+    # @return [Symbol] the endianness of the file, :big or :little
+    attr_reader :endianness
+
     # @return [MachO::MachHeader] if the Mach-O is 32-bit
     # @return [MachO::MachHeader64] if the Mach-O is 64-bit
     attr_reader :header
@@ -281,10 +284,10 @@ module MachO
 
       segment.nsects.times do
         if segment.is_a? SegmentCommand
-          sections << Section.new_from_bin(@raw_data.slice(offset, Section.bytesize))
+          sections << Section.new_from_bin(endianness, @raw_data.slice(offset, Section.bytesize))
           offset += Section.bytesize
         else
-          sections << Section64.new_from_bin(@raw_data.slice(offset, Section64.bytesize))
+          sections << Section64.new_from_bin(endianness, @raw_data.slice(offset, Section64.bytesize))
           offset += Section64.bytesize
         end
       end
@@ -324,7 +327,7 @@ module MachO
 
       magic = get_and_check_magic
       mh_klass = MachO.magic32?(magic) ? MachHeader : MachHeader64
-      mh = mh_klass.new_from_bin(@raw_data[0, mh_klass.bytesize])
+      mh = mh_klass.new_from_bin(endianness, @raw_data[0, mh_klass.bytesize])
 
       check_cputype(mh.cputype)
       check_cpusubtype(mh.cputype, mh.cpusubtype)
@@ -343,6 +346,8 @@ module MachO
 
       raise MagicError.new(magic) unless MachO.magic?(magic)
       raise FatBinaryError.new if MachO.fat_magic?(magic)
+
+      @endianness = MachO.little_magic?(magic) ? :little : :big
 
       magic
     end
@@ -381,7 +386,8 @@ module MachO
       load_commands = []
 
       header.ncmds.times do
-        cmd = @raw_data.slice(offset, 4).unpack("V").first
+        fmt = (endianness == :little) ? "L<" : "L>"
+        cmd = @raw_data.slice(offset, 4).unpack(fmt).first
         cmd_sym = LOAD_COMMANDS[cmd]
 
         raise LoadCommandError.new(cmd) if cmd_sym.nil?
@@ -389,7 +395,7 @@ module MachO
         # why do I do this? i don't like declaring constants below
         # classes, and i need them to resolve...
         klass = MachO.const_get "#{LC_STRUCTURES[cmd_sym]}"
-        command = klass.new_from_bin(@raw_data, offset, @raw_data.slice(offset, klass.bytesize))
+        command = klass.new_from_bin(@raw_data, endianness, offset, @raw_data.slice(offset, klass.bytesize))
 
         load_commands << command
         offset += command.cmdsize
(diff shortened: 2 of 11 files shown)


UniqMartin:
I see you're actively working on this and adjusted your top comment accordingly. Just ping me once you'd like me to have a look! And feel free to do that as an intermediate step, if you're not entirely sure about certain design decisions. (I don't want you to waste time rewriting almost everything, just to discover at the end that the problem might have been solved more elegantly. E.g., peeking at the current diff, I'm not sure if we really want to duplicate all this stuff in `little_format` and `big_format`.)

woodruffw:
@UniqMartin Sure thing!

This is the first solution I've pushed, but I'm experimenting with two others locally. The big problem I've run into is bridging between the class scope (i.e. `new` and `new_from_bin`) and the instance scope (`endianness`, etc). I've hacked between them in various ways, but it ends up being fairly inelegant or significantly more complex than the solution above.

I definitely agree that `little_format` and `big_format` are unnecessarily repetitive - I'm thinking about replacing them with an abstract format that gets translated into the necessary `String#unpack` format.

UniqMartin:
My idea was to mostly stick with the strings being passed into `String#unpack` and slightly adapt what Ruby (2.0, sadly not 1.8.7) has to offer in this regard. Namely, it already offers `L>` for a 32-bit big-endian integer and `L<` for a 32-bit little-endian integer and in my quick hack (that I haven't published anywhere yet) I just used `L=` to denote a 32-bit native-endian integer. Adjusting the format string so that Ruby can handle it could happen in `MachOStructure.new_from_bin` if it gets an additional `endianess` argument.

(Sadly, Ruby 1.8.7 considerably complicates things, as it doesn't have the nice endianess abstraction Ruby 2.0 has to offer and it sorely lacks in 64-bit integer department, as it doesn't provide big- and little-endian format specifiers for them.)

I do wonder whether you'd be interested to look at my undocumented and incomplete (no writing support, no Ruby 1.8.7 support) hack from back then or whether you'd prefer not to be spoiled. (Just because I came up with a solution doesn't necessarily mean that it's a good idea how I did it.)

woodruffw:
> I do wonder whether you'd be interested to look at my undocumented and incomplete (no writing support, no Ruby 1.8.7 support) hack from back then or whether you'd prefer not to be spoiled. (Just because I came up with a solution doesn't necessarily mean that it's a good idea how I did it.)

Sure, I'd definitely like to take a look at it.

I came to roughly the same conclusion re: the newer format specifiers. AFAICT, they're available as early as 1.9.1. Unfortunately (also AFAICT), OS X was pinned to ruby 1.8.7 all the way until 10.8.5. Unless we enable `ruby-macho` only for those later OS X versions, the newer specifiers won't be much help to us.

Another problem that we're going to run into with 1.8.7 (as you mentioned) is the non-existent support for non-system `uint64`s. We can make certain reasonable assumptions about the platform we're running on (namely that it's x86 and therefore `Q` is little-endian), but that means an extra step of converting `Q`-unpacked numbers to big-endian when dealing with PPC Mach-Os. Not the end of the world, but certainly inconvenient.

UniqMartin:
I pushed my endianess work to https://github.com/Homebrew/ruby-macho/compare/master...um/endianess-wip, including a few loosely related additions to load commands and slight changes to CPU subtype handling. (Feel free to pick these up and add them as separate PRs, if you like.) My goal back then was to be able to successfully read all binaries I could get a hold of (`/System/Library`, Homebrew, Xcode, and some older software I had laying around).

As for Ruby 1.8.7, I think it makes sense to focus on Ruby 2.0+. We're all quite unhappy with Ruby 1.8.7 and there's an ongoing effort to create a portable Ruby 2.0 that will run on 10.5+ (possibly even 10.4+) and will soon be auto-installed by Homebrew for those users who are stuck with an old system, thus it's quite possible we can raise the minimum required Ruby version to 2.0 this summer. To be more specific:

- Try to make the code so that it plays nicely with Ruby 2.0 and feels elegant/idiomatic.
- Try to not actively hinder adjustments to make it work with Ruby 1.8.7, but ideally in a way such that 1.8.7 can be added later in a minimally invasive way (and can be removed equally easily once we're happy to raise our minimum to 2.0).
- To that end, modify `.travis.yml` such that:
  - 1.8.7 still runs, but is allowed to fail without failing the complete test run.
  - Maybe add all the major newer Ruby versions (2.1-2.3) to test forward compatibility.

(As usual, just my suggestions. If you have a different plan and other ideas, let's hear them!)

woodruffw:
> As for Ruby 1.8.7, I think it makes sense to focus on Ruby 2.0+. We're all quite unhappy with Ruby 1.8.7 and there's an ongoing effort to create a portable Ruby 2.0 that will run on 10.5+ (possibly even 10.4+) and will soon be auto-installed by Homebrew for those users who are stuck with an old system, thus it's quite possible we can raise the minimum required Ruby version to 2.0 this summer.

This is great news! Having the primary target be 2.x will make this particular task much more pleasant.

I'm going to move some of your improvements into this branch, namely the better `cputype`/`cpusubtype` behavior. I prefer your approach to producing the correct-endian format, and I'm going to try mixing parts of it with my approach. This will drop 1.8.7 compatibility, but could always be hacked back in (if necessary) with a version branch and the re-inclusion of `little_format`/`big_format`.

I'm also going to update `.travis.yml` on master to make 1.8.7 failures allowable, and add 2.1/2.2/2.3 builds.

UniqMartin:
> I'm going to move some of your improvements into this branch, namely the better `cputype` / `cpusubtype` behavior.

I think it is worth refactoring this a bit so that this can be in a separate branch and reviewed (and merged) independently. Ideally, this PR would bring in just the endianess stuff and everything that can be separated out has been already merged by then. And by this I also mean stuff like reorganizing the directory structure in preparation for the PPC/endianess changes.

As for the file/directory names when reshuffling and naming future additions, I'd really like us to standardize on the architecture names Apple uses for all its tools (see `man 3 arch`). This will make things more self-evident for people already familiar with Apple's developer tools. The list of architectures we eventually plan to support should be `i386`, `x86_64`, `ppc`, `ppc64`, `arm`, and `arm64` (from the top of the table in `man 3 arch` and corresponding to the `CPU_SUBTYPE_*_ALL` constants).

> I'm also going to update .travis.yml on master to make 1.8.7 failures allowable, and add 2.1/2.2/2.3 builds.

Rebased my branch on top of this commit and now the build passes (though 1.8.7 fails). :+1:

lib/macho/utils.rb
```diff
@@ -33,4 +33,16 @@ def self.magic32?(num)
   def self.magic64?(num)
     num == MH_MAGIC_64 || num == MH_CIGAM_64
   end
+
+  # @param num [Fixnum] the number being checked
+  # @return [Boolean] true if `num` is a valid little-endian magic number, false otherwise
+  def self.little_magic?(num)
+    num == FAT_CIGAM || num == MH_CIGAM || num == MH_CIGAM_64
```
UniqMartin: You can drop `FAT_CIGAM` here as the `fat_header` structure is *always* big endian, so for that particular structure you can hardcode the endianess as big. (The same applies to the `fat_arch` structure.)

What this also means is that `FAT_CIGAM` should be never used in this Ruby code. Since we read the magic number using the `N` format (32-bit unsigned big-endian integer), the magic number for a fat binary will *always* be `FAT_MAGIC`. This also means that `MachO.magic?` and `MachO.fat_magic?` need to be tweaked a bit. (The `FAT_CIGAM` constant mainly exists for compiled programs, so that they can easily detect whether the binary they are reading has the same endianess or whether all fields need to be byte-swapped during reading.)

lib/macho/utils.rb
```diff
@@ -33,4 +33,16 @@ def self.magic32?(num)
...
+  end
+
+  # @param num [Fixnum] the number being checked
+  # @return [Boolean] true if `num` is a valid big-endian magic number, false otherwise
+  def self.big_magic?(num)
+    num == FAT_MAGIC || num == MH_CIGAM || num == MH_CIGAM_64
```
UniqMartin: You can drop `FAT_MAGIC` here. Basically the same reasons like in `MachO.little_magic?`.

woodruffw:
> I think it is worth refactoring this a bit so that this can be in a separate branch and reviewed (and merged) independently.

Makes sense, I'll move that work to another branch + PR :+1:

lib/macho/utils.rb
woodruffw: Yeah, I realized this while looking at your patch. I'll apply this in a second.

woodruffw:
> As for the file/directory names when reshuffling and naming future additions, I'd really like us to standardize on the architecture names Apple uses for all its tools (see man 3 arch). This will make things more self-evident for people already familiar with Apple's developer tools. The list of architectures we eventually plan to support should be i386, x86_64, ppc, ppc64, arm, and arm64 (from the top of the table in man 3 arch and corresponding to the CPU_SUBTYPE_*_ALL constants).

This sounds good to me. Do you have any ideas w/r/t categorizing fat binaries? I went with 'x86' for that reason, but I'm sure we can come up with a better naming scheme for identifying binaries with multiple architectures.

.gitignore
```diff
@@ -16,3 +16,6 @@
 /test/src/*.dylib
 /test/src/*.o
 /test/src/*.so
+
+# Do NOT ignore binaries placed in subdirs under '/test/bin'.
+!/test/bin/*/*
```
UniqMartin: Why is this needed? I can't seem to find the rule(s) that would cause the files that are un-ignored here to be ignored in the first place.

woodruffw: I added it late last night without thinking, just realized that my global `.gitignore` was the source of my troubles. I'll undo that.

test/helpers.rb
```diff
@@ -4,17 +4,17 @@
...
+  TEST_EXE = "test/bin/x86/hello.bin"
+  TEST_DYLIB = "test/bin/x86/libhello.dylib"
+  TEST_EXTRA_DYLIB = "test/bin/x86/libextrahello.dylib"
+  TEST_BUNDLE = "test/bin/x86/hellobundle.so"
+
+  TEST_FAT_OBJ = "test/bin/x86/fathello.o"
+  TEST_FAT_EXE = "test/bin/x86/fathello.bin"
+  TEST_FAT_DYLIB = "test/bin/x86/libfathello.dylib"
+  TEST_FAT_EXTRA_DYLIB = "test/bin/x86/libfatextrahello.dylib"
+  TEST_FAT_BUNDLE = "test/bin/x86/fathellobundle.so"
```
UniqMartin: Maybe rename the constants and use `i386` or `x86_64` instead of of `x86` for consistency with `man 3 arch` as pointed out in https://github.com/Homebrew/ruby-macho/pull/24#issuecomment-223326092. For fat binaries I'd even suggest to put them in a separate directory, e.g. `fat-i386-ppc` would be used for a classic universal binary. `fat-i386-x86_64` would be used for a more modern 32/64-bit Intel universal binary.

Another idea: Drop all the constants and instead use a helper method that constructs the path for you, e.g.:

```ruby
def binary_fixture(archs, name)
  arch_dir = archs.is_a?(Array) ? "fat-#{archs.join("-")}" : archs.to_s
  "test/bin/#{arch_dir}/#{name}"
end
```

It would be used something like this:

```ruby
binary_fixture(:i386, "hello.o") # Yields `test/bin/i386/hello.o`.
binary_fixture([:i386, :ppc], "libhello.dylib") # Yields `test/bin/fat-i386-ppc/libhello.dylib`.
```

UniqMartin:
I left a few code comments that already made sense at this point. (And I notice you are pretty quick to follow up. 😄)

> Do you have any ideas w/r/t categorizing fat binaries? I went with 'x86' for that reason, but I'm sure we can come up with a better naming scheme for identifying binaries with multiple architectures.

See this code comment: https://github.com/Homebrew/ruby-macho/pull/24#discussion_r65569062.

.gitignore
UniqMartin: I suspected something like that. No worries, it's no big deal! It just caught my attention. :wink:

woodruffw:
> Another idea: Drop all the constants and instead use a helper method that constructs the path for you

This makes the most sense to me, so I'll go with it.

These changes will probably involve refactoring a lot of the test code, so I'll put it in another PR.

UniqMartin:
> These changes will probably involve refactoring a lot of the test code, so I'll put it in another PR.

:+1: Let's do some cleanup of the test code and CPU (sub)type handling and some rearrangement of the test binaries first and once that's done, rebase this endianess PR on top of these changes.

woodruffw:
> Let's do some cleanup of the test code and CPU (sub)type handling and some rearrangement of the test binaries first and once that's done, rebase this endianess PR on top of these changes.

Sounds great!

woodruffw:
Took the test-specific code out of this PR, the general changes you proposed are now in #26 :tada:

lib/macho/utils.rb
```diff
@@ -19,7 +19,7 @@ def self.magic?(num)
   # @param num [Fixnum] the number being checked
   # @return [Boolean] true if `num` is a valid Fat magic number, false otherwise
   def self.fat_magic?(num)
-    num == FAT_MAGIC || num == FAT_CIGAM
+    num == FAT_MAGIC
```
UniqMartin: Maybe it's best to document why `FAT_CIGAM` is never considered (here, next to where the constant is defined, or even in both locations). You'll also need to drop `FAT_CIGAM` from the `MH_MAGICS` hash to make all the `*magic?` methods consistent.

woodruffw: Oop, missed that. I'll document it's non-inclusion where it's defined.

woodruffw:
Now that #25 and #26 are merged (and we're rebased here), I'll go ahead on here :tada:

woodruffw:
9a50914 contains a mixture of your solution and some of the changes I made earlier. I think it's a pretty decent one (better than my two-variable one, and not nearly as invasive as the ones I planned earlier).

UniqMartin:
Sorry for the delay here, a proper review will take me a bit more time and it's already too late where I am. In the mean time, can you please:

- Consolidate the existing commits by interactively rebasing them.

- Try to make the load command specifications more concise. All those `self.format =` and `self.sizeof =` don't look very nice to me. Would you mind reverting that back to the `FORMAT` and `SIZEOF` constants used before or is there some advantage to the new style I have overlooked?

  Long-term (certainly not in this PR), I think I'd like us to have a mini-DSL to define those structures. It could look something like this:

  ```ruby
  class SegmentCommand64 < LoadCommand
    field :segname, :string, :length => 16
    field :vmaddr, :uint64
    field :vmsize, :uint64
    field :fileoff, :uint64
    field :filesize, :uint64
    field :maxprot, :int32
    field :initprot, :int32
    field :nsects, :uint32
    field :flags, :uint32

    # …
  end
  ```

  (This would implicitly set up `format` and `sizeof` and would also take inheritance into account, so the fields from `LoadCommand` don't need to be repeated here.)

- Look into setting up the corresponding tests and prepare 32/64-bit PPC binaries.

woodruffw:
> Consolidate the existing commits by interactively rebasing them.

Sure thing!

> Try to make the load command specifications more concise. All those self.format = and self.sizeof = don't look very nice to me. Would you mind reverting that back to the FORMAT and SIZEOF constants used before or is there some advantage to the new style I have overlooked?

My thought in doing this was that it's slightly clearer as to the inheritance relationship between various load commands, `LoadCommand`, and `MachOStructure`, but it's mostly a side effect of other approaches I was trying. I'll revert it.

> Long-term (certainly not in this PR), I think I'd like us to have a mini-DSL to define those structures.

I think that's a great idea!

> Look into setting up the corresponding tests and prepare 32/64-bit PPC binaries.

Will do. I'd prefer that we integrate the PPC tests in a separate PR, but it's your call :+1:

lib/macho/load_commands.rb
```diff
@@ -157,16 +157,18 @@ class LoadCommand < MachOStructure
...
     SIZEOF = 8
 
     # Creates a new LoadCommand given an offset and binary string
     # @param offset [Fixnum] the offset to initialize with
     # @param bin [String] the binary string to initialize with
     # @return [MachO::LoadCommand] the new load command
     # @api private
-    def self.new_from_bin(raw_data, offset, bin)
-      self.new(raw_data, offset, *bin.unpack(self::FORMAT))
+    def self.new_from_bin(raw_data, endianness, offset, bin)
```
UniqMartin: This lacks documentation of the pre-existing `raw_data` argument and the newly added `endianess` argument.

lib/macho/macho_file.rb
```diff
@@ -8,6 +8,8 @@ class MachOFile
     # @return [String] the filename loaded from, or nil if loaded from a binary string
     attr_accessor :filename
 
+    attr_reader :endianness
```
UniqMartin: Looks like it's missing some documentation. :wink:

lib/macho/macho_file.rb
```diff
@@ -381,15 +385,16 @@ def get_load_commands
...
+        cmd = @raw_data.slice(offset, 4).unpack(fmt).first
         cmd_sym = LOAD_COMMANDS[cmd]
 
         raise LoadCommandError.new(cmd) if cmd_sym.nil?
 
         # why do I do this? i don't like declaring constants below
         # classes, and i need them to resolve...
         klass = MachO.const_get "#{LC_STRUCTURES[cmd_sym]}"
-        command = klass.new_from_bin(@raw_data, offset, @raw_data.slice(offset, klass.bytesize))
+        command = klass.new_from_bin(@raw_data, @endianness, offset, @raw_data.slice(offset, klass.bytesize))
```
UniqMartin: I think for read access we want to rely on the `endianess` reader method instead of the `@endianess` instance variable.

lib/macho/structure.rb
```diff
@@ -2,11 +2,11 @@ module MachO
   # A general purpose pseudo-structure.
   # @abstract
   class MachOStructure
-    # The format of the data structure, in String#unpack format.
+    # The String#unpack format of the data structure.
     FORMAT = ""
 
     # The size of the data structure, in bytes.
-    SIZEOF = 0
+    SIZEOF = ""
```
UniqMartin: This looks wrong.

lib/macho/structure.rb
```diff
@@ -15,8 +15,22 @@ def self.bytesize
 
     # @return [MachO::MachOStructure] a new MachOStructure initialized with `bin`
     # @api private
-    def self.new_from_bin(bin)
-      self.new(*bin.unpack(self::FORMAT))
+    def self.new_from_bin(endianness, bin)
```
UniqMartin: You might want to document the new `endianess` argument and the pre-existing but undocumented `bin` argument.

woodruffw:
All silly mistakes/oversights on my part, thanks!

UniqMartin:
Some minor code comments, but we're getting closer to merging this. :+1:

>> Consolidate the existing commits by interactively rebasing them.
> 
> Sure thing!

Didn't expect you to squash everything into a single commit, just the slight back and forth between different implementations of basically the same change. But if you're happy with it, I'm fine with it.

*Side note: If a PR happens to consist of only a single commit or is small enough to be squashed down to a single commit without losing valuable history, consider using the relatively new option of performing a [squash merge](https://github.com/blog/2141-squash-your-commits) directly from the GitHub website. (Notice the triangle on the right of the button once you hit “Merge pull request” and are asked to confirm.) This keeps the history a bit tidier by reserving merge commits to branches with more substantial changes consisting of several commits.*

> I'd prefer that we integrate the PPC tests in a separate PR, but it's your call 👍

I'm not feeling too strongly about this, but I would be more confident if we had *some* tests that show that the code can correctly extract some basic information from big-endian binaries. (I'm not asking for an exhaustive test suite here. Even for `i386`/`x86_64` we're quite a bit away from having one.)

Do you already have a plan for generating (or obtaining) various PPC binaries for testing purposes? Do you need help with that? (I believe support for generating PPC code has been removed from Apple's compilers quite a while ago.)

woodruffw:
> Didn't expect you to squash everything into a single commit, just the slight back and forth between different implementations of basically the same change. But if you're happy with it, I'm fine with it.

Oh, that was an oversight. Oops. I have an almost identical local branch for the earlier implementation, so nothing was lost.



> Do you already have a plan for generating (or obtaining) various PPC binaries for testing purposes? Do you need help with that? (I believe support for generating PPC code has been removed from Apple's compilers quite a while ago.)

I actually have an iMac G4 sitting next to me. I haven't plugged it in in nearly a decade, but (if it still works) I could use it to churn out the PPC binaries we need. I'll try it right now.

UniqMartin:
> I actually have an iMac G4 sitting next to me. I haven't plugged it in in nearly a decade, but (if it still works) I could use it to churn out the PPC binaries we need. I'll try it right now.

I'm a bit amazed you have access to something like this, but that's of course super helpful for this project. I started using Apple hardware with OS X 10.5, thus after the transition from PowerPC to Intel …

woodruffw:
I think it was my family's very first computer, I've had it sitting in a closet ever since 10.5. Hopefully I can remember the password, but I'm glad it's finally being put to use again :smile:

woodruffw:
I haven't remembered the administrator account's password yet, but I scraped a few PPC binaries from the disk on another account. I'll use them to test this branch (and I can upload them if you'd like).

woodruffw:
@UniqMartin Do you have any fat (PPC) binaries that I could test with?

UniqMartin:
Sorry for the lack of feedback here. I hadn't realized you pushed changed to this branch a few days ago.

> Do you have any fat (PPC) binaries that I could test with?

I believe I have. Please bear with me until tomorrow, when I will be less tired. Then I'll dig up some stuff and provide proper feedback on what you added since I last had a chance to look at this PR.

woodruffw:
> Sorry for the lack of feedback here

No worries :smile:

UniqMartin:
Sorry for some more delay! Took me a while to get my hands at an OS X 10.6.8 VM with Xcode 3.2.6, so I could produce the usual test binaries with a modified version of `test/src/Makefile`. Not sure what the exact origin of the PPC binaries is, but if you haven't compiled them yourself I'm a bit hesitant to include binary code—however trivial—sourced from a third party with unclear licensing.

If you're fine with that, my proposal would be that I merge https://github.com/Homebrew/ruby-macho/compare/master...um/test-makefile and then you can rebase this PR on top of that and adjust a few file names. I only included the interesting cases (`ppc` and `fat-i386-ppc`) in my commit, but if you would like to take a closer look, [xcode-3.2.6.zip](https://github.com/Homebrew/ruby-macho/files/316932/xcode-3.2.6.zip) has the full set of binaries. Some differences from the already existing binaries:

- Use `-weak_library` instead of `-upward_library`. `ld` from Xcode 3.2.6 doesn't support the latter.
- Use `-reexport-l` instead of `-lazy-l`. `ld` from Xcode 3.2.6 supports the latter, but gets confused.
- Use `/usr/lib/libz.dylib` instead of `/usr/lib/libm.dylib` because the former exists on all OS X releases I've ever encountered (unlike the latter).

(All binaries were built on the mentioned VM with `make USE=10.6-xcode3.2.6`.)

test/test_macho.rb
```diff
@@ -28,262 +28,312 @@ def test_truncated_file
...
-      assert_kind_of Fixnum, lc.offset
-      assert_kind_of Fixnum, lc.cmd
-      assert_kind_of Fixnum, lc.cmdsize
-      assert_kind_of String, lc.to_s
-      assert_kind_of Symbol, lc.type
-      assert_kind_of Symbol, lc.to_sym
+    files = [
+      MachO::MachOFile.new(fixture(:x86_64, "hello.bin")),
+      MachO::MachOFile.new(fixture(:ppc, "echo.bin"))
+    ]
```
UniqMartin: I'd prefer for this array to only contain the file names (`fixture(…)`) and the `MachOFile` instances be created inside the block of the following `each do`.

UniqMartin: (Obviously also applies to all other instances of this pattern.)

test/test_macho.rb
```diff
@@ -28,262 +28,312 @@ def test_truncated_file
...
-    assert_equal "testing", file.dylib_id
-
-    # change it back within the same instance
-    file.dylib_id = old_id
-    assert_equal old_id, file.dylib_id
+    groups = [
+      [
+        fixture(:x86_64, "libhello.dylib"),
+        fixture(:x86_64, "libhello_actual.dylib"),
+        expected = fixture(:x86_64, "libhello_expected.dylib")
```
UniqMartin: The `expected =` needs to be dropped here.

woodruffw:
> Not sure what the exact origin of the PPC binaries is, but if you haven't compiled them yourself I'm a bit hesitant to include binary code—however trivial—sourced from a third party with unclear licensing.

They're all open, but having all the binaries be from a uniform source is obviously preferable, so I'll get rid of these post-rebase :+1: 

> my proposal would be that I merge master...um/test-makefile and then you can rebase this PR on top of that and adjust a few file names

This sounds great to me, merge at your convenience.

> Use -weak_library instead of -upward_library. ld from Xcode 3.2.6 doesn't support the latter.
> Use -reexport-l instead of -lazy-l. ld from Xcode 3.2.6 supports the latter, but gets confused.
> Use /usr/lib/libz.dylib instead of /usr/lib/libm.dylib because the former exists on all OS X releases I've ever encountered (unlike the latter).

Noted, I'll remember these when reexamining the tests.

test/test_macho.rb
woodruffw: Got it!

UniqMartin:
Alright, merged the binaries to `master` a few moments ago. Aside from this I have only two minor code comments. This PR is almost ready to be merged … :tada:

test/test_macho.rb
woodruffw: Whoops! I'll fix that.

UniqMartin:
Not sure if this is just transient, but the `git rebase` didn't quite turn out as it should have. It looks like you cherry-picked the two commits I added to `master` into your branch instead of rebasing your branch on top of a the new `master`. (Feel free to ping me on Slack if I can help you with cleaning this up.)

UniqMartin:
If you're happy with this PR, I'm too. Feel free to 🚢 this! I think it went through enough iterations and refinements to be ready and we can always make adjustments in follow-up PRs, if necessary.

UniqMartin:
:+1: :tada:

PR #25, opened 2016-06-02
Improve cputype/cpusubtype handing.

woodruffw:
Additionally, adds new CPU types and subtypes, as well as two
additional load commands and freezes declared data structures.

@UniqMartin These are mostly your changes, plus some of the constants I added in #24.

### DIFF
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -49,17 +49,18 @@ module MachO
 
   # Raised when the CPU type is unknown.
   class CPUTypeError < MachOError
-    # @param num [Fixnum] the unknown number
-    def initialize(num)
-      super "Unrecognized CPU type: 0x#{"%02x" % num}"
+    # @param cputype [Fixnum] the unknown CPU type
+    def initialize(cputype)
+      super "Unrecognized CPU type: 0x#{"%08x" % cputype}"
     end
   end
 
-  # Raised when the CPU subtype is unknown.
+  # Raised when the CPU type/sub-type pair is unknown.
   class CPUSubtypeError < MachOError
-    # @param num [Fixnum] the unknown number
-    def initialize(num)
-      super "Unrecognized CPU sub-type: 0x#{"%02x" % num}"
+    # @param cputype [Fixnum] the CPU type of the unknown pair
+    # @param cpusubtype [Fixnum] the CPU sub-type of the unknown pair
+    def initialize(cputype, cpusubtype)
+      super "Unrecognized CPU sub-type: 0x#{"%08x" % cpusubtype} (for CPU type: 0x#{"%08x" % cputype})"
     end
   end
 
--- a/lib/macho/headers.rb
+++ b/lib/macho/headers.rb
@@ -39,19 +39,27 @@ module MachO
   # x86_64 (AMD64) compatible CPUs
   CPU_TYPE_X86_64 = (CPU_TYPE_I386 | CPU_ARCH_ABI64)
 
-  # PowerPC compatible CPUs (7400 series?)
+  # 32-bit ARM compatible CPUs
+  CPU_TYPE_ARM = 0x0c
+
+  # 64-bit ARM compatible CPUs
+  CPU_TYPE_ARM64 = (CPU_TYPE_ARM | CPU_ARCH_ABI64)
+
+  # PowerPC compatible CPUs
   CPU_TYPE_POWERPC = 0x12
 
-  # PowerPC64 compatible CPUs (970 series?)
+  # PowerPC64 compatible CPUs
   CPU_TYPE_POWERPC64 = (CPU_TYPE_POWERPC | CPU_ARCH_ABI64)
 
-  # association of cpu types to string representations
+  # association of cpu types to symbol representations
   CPU_TYPES = {
-    CPU_TYPE_ANY => "CPU_TYPE_ANY",
-    CPU_TYPE_I386 => "CPU_TYPE_I386",
-    CPU_TYPE_X86_64 => "CPU_TYPE_X86_64",
-    CPU_TYPE_POWERPC => "CPU_TYPE_POWERPC",
-    CPU_TYPE_POWERPC64 => "CPU_TYPE_POWERPC64"
+    CPU_TYPE_ANY => :any,
+    CPU_TYPE_I386 => :i386,
+    CPU_TYPE_X86_64 => :x86_64,
+    CPU_TYPE_ARM => :arm,
+    CPU_TYPE_ARM64 => :arm64,
+    CPU_TYPE_POWERPC => :ppc,
+    CPU_TYPE_POWERPC64 => :ppc64,
   }
 
   # mask for CPU subtype capabilities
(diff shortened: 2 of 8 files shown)


lib/macho/headers.rb
```diff
@@ -50,6 +56,8 @@ module MachO
     CPU_TYPE_ANY => "CPU_TYPE_ANY",
     CPU_TYPE_I386 => "CPU_TYPE_I386",
     CPU_TYPE_X86_64 => "CPU_TYPE_X86_64",
+    CPU_TYPE_ARM => "CPU_TYPE_ARM",
+    CPU_TYPE_ARM64 => "CPU_TYPE_ARM64",
```
UniqMartin: I'm not sure how you feel about this, especially since this has already been exposed publicly, but I'd like us to consider changing the right hand side from a string that is identical to the constant name (both ugly and unwieldy when exposed to a user or in the `ruby-macho` API) to a shorter symbol. Again, I'm proposing to use the established identifiers from `man 3 arch`.

lib/macho/headers.rb
```diff
@@ -61,17 +69,57 @@ module MachO
...
+    [CPU_TYPE_ARM, CPU_SUBTYPE_ARM_V7] => "CPU_SUBTYPE_ARM_V7",
+    [CPU_TYPE_ARM, CPU_SUBTYPE_ARM_V7S] => "CPU_SUBTYPE_ARM_V7S",
+    [CPU_TYPE_ARM, CPU_SUBTYPE_ARM_V7K] => "CPU_SUBTYPE_ARM_V7K",
+    [CPU_TYPE_ARM64, CPU_SUBTYPE_ARM64_ALL] => "CPU_SUBTYPE_ARM64_ALL",
+    [CPU_TYPE_POWERPC, CPU_SUBTYPE_POWERPC_750] => "CPU_SUBTYPE_POWERPC_750",
+    [CPU_TYPE_POWERPC, CPU_SUBTYPE_POWERPC_7400] => "CPU_SUBTYPE_POWERPC_7400",
+    [CPU_TYPE_POWERPC, CPU_SUBTYPE_POWERPC_7450] => "CPU_SUBTYPE_POWERPC_7450",
+    [CPU_TYPE_POWERPC, CPU_SUBTYPE_POWERPC_ALL] => "CPU_SUBTYPE_POWERPC_ALL",
+    [CPU_TYPE_POWERPC64, CPU_SUBTYPE_POWERPC_970] => "CPU_SUBTYPE_POWERPC_970",
+    [CPU_TYPE_POWERPC64, CPU_SUBTYPE_POWERPC64_ALL] => "CPU_SUBTYPE_POWERPC64_ALL"
```
UniqMartin: Same story like with the CPU type identifiers, i.e. make them shorter, symbols, and consistent with `man 3 arch`. Also, on second thought I think it's cleaner and avoids repetition when we make this a two-level hierarchy of hashes, e.g.

```ruby
CPU_SUBTYPES = {
  CPU_TYPE_ARM = {
    CPU_SUBTYPE_ARM_V7 => :armv7,
    CPU_SUBTYPE_ARM_V7S => :armv7s,
    # ...
  }.freeze,
  # ...
}.freeze
```

lib/macho/load_commands.rb
```diff
@@ -52,8 +52,10 @@ module MachO
     0x2b => :LC_DYLIB_CODE_SIGN_DRS,
     0x2c => :LC_ENCRYPTION_INFO_64,
     0x2d => :LC_LINKER_OPTION,
-    0x2e => :LC_LINKER_OPTIMIZATION_HINT
-  }
+    0x2e => :LC_LINKER_OPTIMIZATION_HINT,
+    0x2f => :LC_VERSION_MIN_TVOS,
+    0x30 => :LC_VERSION_MIN_WATCHOS
```
UniqMartin: Style nit: Always use a comma here (and in other hashes/arrays) to terminate the last element. Makes both the diff cleaner if things are added in the future and makes reordering elements less error-prone (because the last element is no different from every other element).

lib/macho/headers.rb
woodruffw: My main motivation for leaving it like that was matching `libmacho`'s constant names, since those are used throughout `otool`.

I agree that they're very ugly and unwieldy, and I think that replacing them with shorthands is a good idea. Backwards compatibility shouldn't be a *huge* problem, as the library is still fairly immature and had broken compatibility in some aspect with virtually every release.

I'll get to work on this :+1:

lib/macho/headers.rb
woodruffw: This looks good to me as well. A two-level hashmap is much cleaner than array-pairs in this case.

woodruffw: Since this covers the entirety of all CPU types and subtypes, I'm just going to get rid of get rid of `CPU_TYPES` and `CPU_SUBTYPES` in favor of `CPU_TYPE_TREE`.

lib/macho/load_commands.rb
woodruffw: Yeah, I have a bad habit of not doing that. I'll re-add them as I continue to change the relevant parts of these files.

lib/macho/headers.rb
```diff
@@ -61,17 +69,69 @@ module MachO
...
+  CPU_SUBTYPE_POWERPC_7400 = 10
+
+  # PowerPC G4 "Voyager"
+  CPU_SUBTYPE_POWERPC_7450 = 11
+
+  # PowerPC G5
+  CPU_SUBTYPE_POWERPC_970 = 100
+
+  # all CPU sub-types for CPU type `CPU_TYPE_POWERPC`
+  CPU_SUBTYPE_POWERPC_ALL = 0
```
UniqMartin: The comment above this doesn't really apply since you've added the corresponding sub-types in separate “sections” above this block. Not sure if they deserve this much room. Maybe put them all below this line and either omit the comment about the corresponding model or put that comment at the end of the line?

lib/macho/headers.rb
```diff
@@ -61,17 +69,69 @@ module MachO
...
+      CPU_SUBTYPE_ARM64_ALL => :arm64,
+    }.freeze,
+    CPU_TYPE_POWERPC => {
+      CPU_SUBTYPE_POWERPC_750 => :ppc750,
+      CPU_SUBTYPE_POWERPC_7400 => :ppc7400,
+      CPU_SUBTYPE_POWERPC_7450 => :ppc7450,
+      CPU_SUBTYPE_POWERPC_ALL => :ppc,
+    }.freeze,
+    CPU_TYPE_POWERPC64 => {
+      CPU_SUBTYPE_POWERPC_970 => :ppc970,
```
UniqMartin: I believe this should be in the `CPU_TYPE_POWERPC` sub-hash.

woodruffw: I was conflicted about this one, as the 970 (G5) was definitely 64-bit, but I'm unclear as to how many binaries were built solely for it. If we move it into `CPU_TYPE_POWERPC`, is it inconsistent to have `CPU_TYPE_ARM` and `CPU_TYPE_ARM64` be distinct?

lib/macho/headers.rb
woodruffw: I think moving them all below `POWERPC{,64}_ALL` and specifying the comment as "any" rather than "all" CPU sub-types would be the most clear :+1:

UniqMartin:
Currently, an unknown CPU sub-type is a hard failure. I'd like us to reconsider this decision. I think we need two modes of operation and maybe you can think a bit about how to best implement this (and switch between these modes). These are the scenarios I have in mind:

- For Homebrew it would be very unfortunate if a new CPU sub-type appears and the library starts failing and throwing exceptions, despite that it fully understands the underlying CPU type and there's nothing really sub-type specific in the code. Returning some auto-generated identifier, e.g. `:x86_64_unknown`, as a fallback is probably best.

- For testing purposes and maybe also for some other use cases it makes sense to throw an exception (and not just print a warning) to immediately notice the omission (or bug) in the sub-type assignment and to fix/extend the library accordingly.

lib/macho/headers.rb
UniqMartin: I think you're a bit confused here. The PowerPC 970 CPU supported both 32-bit and 64-bit binaries.

32-bit binaries were, if they relied on CPU instruction unavailable prior to the 970 got a type of `CPU_TYPE_POWERPC` and a sub-type of `CPU_SUBTYPE_POWERPC_970`. But the 970 was still perfectly capable of executing 32-bit binaries with other (earlier introduced) sub-types.

For 64-bit binaries a type of `CPU_TYPE_POWERPC64` and sub-type of `CPU_SUBTYPE_POWERPC64_ALL` effectively means PowerPC 970, because this was the first 64-bit PowerPC processor. (If there had been any successors, they most likely would have been compatible to the 970 and just extended its instruction set, possibly making other `CPU_SUBTYPE_POWERPC64_*` sub-types necessary.)

woodruffw:
> For Homebrew it would be very unfortunate if a new CPU sub-type appears and the library starts failing and throwing exceptions, despite that it fully understands the underlying CPU type and there's nothing really sub-type specific in the code. Returning some auto-generated identifier, e.g. :x86_64_unknown, as a fallback is probably best.

Since Homebrew needs to play catch-up with OS X with every successive release anyway, is this really going to be a problem? As I see it currently, throwing an exception for a new CPU sub-type will help us detect changes to Mach-O earlier (since Apple isn't always timely about releasing updated sources) and mitigate more serious changes to the spec (like the kind between 10.4/10.5/10.6).

I understand that we want to make the UX as resilient to interruptions from the library as possible, but I see `ruby-macho` being updated per-OS X-release like the rest of Homebrew. If assumption is correct (and it very well may not be :smile:), then hiding little errors like unknown sub-types may put us in a situation where users try Homebrew on a new OS X release and `ruby-macho` silently does the wrong thing because Apple has changed the playing field.

lib/macho/headers.rb
woodruffw: Alright, I think I see what you mean. If I'm understanding correctly, it should be like this, then?

```ruby
# ...
CPU_TYPE_POWERPC => {
  # ...
  CPU_SUBTYPE_POWERPC_970 => :ppc970
},
CPU_TYPE_POWERPC64 => {
  CPU_SUBTYPE_POWERPC64_ALL => :ppc64
}
# ...
```

lib/macho/headers.rb
UniqMartin: The thing is that these always come in pairs and refer to certain architectures and instruction sets, not necessarily specific processors (of which some are capable of both 32-bit and 64-bit modes). Thus there's always one `CPU_TYPE_<type>` constant and at least one (but usually many) `CPU_SUBTYPE_<type>_<subtype>` constants strictly belonging to it. The sub-type field simply cannot be interpreted independently of the type field.

lib/macho/headers.rb
UniqMartin: Exactly. And both these sub-types mean: I need at least a PowerPC 970 CPU to run. Which one applies depends on whether the binary is 32-bit or 64-bit.

woodruffw: Sounds good. Yeah, I was a bit confused :confounded:

UniqMartin:
Of course there will be a certain degree of catch-up and I'm somewhat fine with, e.g. Homebrew's CI throwing those exceptions when it comes across a previously unsupported Mach-O binary. I would personally prefer for the library to be more resilient when used by a regular Homebrew user and I honestly don't think that breaking a user's setup is warranted just because we didn't add a new constant in time (but didn't need to change anything else). Note that I'm explicitly only talking about CPU sub-types, unknown CPU types should still cause a failure.

But that's just my opinion and maybe it's best to ask other maintainers how they feel about this. In any case, that's not an issue that is blocking this PR. A change to the failure behavior, if we deem it necessary, can happen later and separately. For now let's focus on getting a solid foundation for the endianess work that is our first major milestone this summer.

woodruffw:
> But that's just my opinion and maybe it's best to ask other maintainers how they feel about this. In any case, that's not an issue that is blocking this PR. A change to the failure behavior, if we deem it necessary, can happen later and separately. For now let's focus on getting a solid foundation for the endianess work that is our first major milestone this summer.

Sounds good to me. I'm definitely not opposed to changes of this type, so I'll give some thought to accomplishing this :smile:.

lib/macho/headers.rb
woodruffw: Yeah, I think I understand that now - the confusion was mostly self-imposed. I'll move `POWERPC64_SUBTYPE_ALL` back.

woodruffw:
Just a FYI, I'll squash this down to 1-2 commits when we're ready to merge.

lib/macho/exceptions.rb
```diff
@@ -49,17 +49,18 @@ def initialize
...
     end
   end
 
-  # Raised when the CPU subtype is unknown.
+  # Raised when the CPU type/sub-type pair is unknown.
   class CPUSubtypeError < MachOError
-    # @param num [Fixnum] the unknown number
+    # @param cputype [Fixnum] the CPU type of the unknown pair
+    # @param cpusubtype [Fixnum] the CPU sub-type of the unknown pair
     def initialize(num)
```
UniqMartin: The method signature hasn't caught up with the changes. :wink:

woodruffw: Whoops! Fixed.

lib/macho/headers.rb
```diff
@@ -61,17 +69,69 @@ module MachO
...
+  # all CPU sub-types for CPU type `CPU_TYPE_I386`
   CPU_SUBTYPE_X86_ALL = 3
 
-  # all x86-type CPUs (what makes this different from CPU_SUBTYPE_X86_ALL?)
-  CPU_SUBTYPE_X86_ARCH1 = 4
+  # all CPU sub-types for CPU type `CPU_TYPE_X86_64`
+  CPU_SUBTYPE_X86_64_ALL = CPU_SUBTYPE_X86_ALL
+
+  # all CPU sub-types for CPU type `CPU_TYPE_X86_64` (Haskell tree)
+  CPU_SUBTYPE_X86_64_H = 8
```
UniqMartin: Drop the comment and move this closer to the constant above?

woodruffw: I had them separate because YARD only documents them correctly when separated. That was a personal preference, so I'm happy to change it if you prefer.

UniqMartin:
Would you mind addressing https://github.com/Homebrew/ruby-macho/pull/25#discussion_r65572540 (only in the changed code) and https://github.com/Homebrew/ruby-macho/pull/25#discussion_r65736119 (which I believe you intended to address, but didn't yet)? Other than that I'm happy with this PR.

lib/macho/headers.rb
UniqMartin: Oh, I see. I didn't consider documentation with YARD. If things look nicer like this in the YARD documentation, let's keep them.

woodruffw:
Added the trailing commas.

Re: the other comment: Does the "any" instead of "all" (changed in aafb80dec8b0e6d871c977e31ca9f14706763e3a) make it any clearer? I'm not really sure how to document a constant like `CPU_SUBTYPE_POWERPC_ALL`.

UniqMartin:
Should be good to go. Feel free to merge as-is!

> Re: the other comment: Does the "any" instead of "all" (changed in aafb80dec8b0e6d871c977e31ca9f14706763e3a) make it any clearer? I'm not really sure how to document a constant like `CPU_SUBTYPE_POWERPC_ALL`.

It's definitely an improvement, though I'm not 100% happy with the “any”. But I can't currently think of a better wording. For most architectures it's something like a generic CPU or the lowest common denominator or something like that …

woodruffw:
> It's definitely an improvement, though I'm not 100% happy with the “any”

Yeah, me neither. I think "lowest common denominator" gets imparts the message slightly better, so I'll change it to that and merge.

woodruffw:
Added the 'all' subtype for ARM and V8 for ARM/ARM64 for good measure as well.

UniqMartin:
:+1:

woodruffw:
Alright! I'll just squash this down and merge it now.

PR #26, opened 2016-06-02
test: Restructure bin folder, replace hardcoded paths with generator.

woodruffw:
Moves binaries to subdirectories under bin/ corresponding to their
architecture (following `man 3 arch`). Fat binaries are placed in directories
of the schema "fat-arch1-arch2-...-archN". Paths to appropriate binaries
are then produced by the Helper#fixture method, which takes a list of
architectures and a filename to stitch into a path.

cc @UniqMartin


UniqMartin:
Glad you liked my idea for organizing these. 😸 You might want to adjust the [`Makefile` in `test/src/`](https://github.com/Homebrew/ruby-macho/blob/6d7f1294b868283a1dd615bd75fa69b6a1320509/test/src/Makefile) to match the file names and to modify it slightly such that it directly drops the produced binaries in the `test/bin/<arch>/` directories. But other than that, looks good to me.

woodruffw:
> You might want to adjust the Makefile in test/src/ to match the file names and to modify it slightly such that it directly drops the produced binaries in the test/bin/<arch>/

Good idea, I'll add that to this PR :+1:

woodruffw:
The last commit is still WIP since I don't have my Mac to test the Makefile with.

UniqMartin:
`Makefile` works for me (just tested locally), but the `.gitignore` needs some minor adjustment:

```diff
diff --git i/.gitignore w/.gitignore
index 26c3f4bc..cf599042 100644
--- i/.gitignore
+++ w/.gitignore
@@ -11,8 +11,8 @@
 /vendor/bundle/

 # Build results for test cases before they are selectively moved to '/test/bin'.
-/test/src/*.a
-/test/src/*.bin
-/test/src/*.dylib
-/test/src/*.o
-/test/src/*.so
+/test/src/*/*.a
+/test/src/*/*.bin
+/test/src/*/*.dylib
+/test/src/*/*.o
+/test/src/*/*.so
```

Feel free to merge after including this change.

PR #27, opened 2016-06-16
headers: Add additional ARM subtypes, correct errors.

woodruffw:
@UniqMartin How much completeness do you think we should aim for re: CPU types and subtypes?
I'm thinking about filling out every group that appears in [arch.c](https://opensource.apple.com/source/cctools/cctools-877.8/libstuff/arch.c), as these are the ones that certainly had binaries built (or at least targets available) at some point. We probably won't ever run into an HPPA Mach-O, but it can't hurt to parse one correctly at no additional cost.

### DIFF
--- a/lib/macho.rb
+++ b/lib/macho.rb
@@ -12,5 +12,5 @@ require "#{File.dirname(__FILE__)}/macho/tools"
 # The primary namespace for ruby-macho.
 module MachO
   # release version
-  VERSION = "0.2.3".freeze
+  VERSION = "0.2.4".freeze
 end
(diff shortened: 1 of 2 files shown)


UniqMartin:
I agree! I think it's a nice goal to aim for completeness in this area and adding a few more constants and symbols to the code doesn't really cost us much, but we benefit from much more graceful handling (and a meaningful identifier) if we ever come across such binaries. (This makes me wonder if there's some online resource that collects ancient Mach-O binaries.)

Code looks fine to me, thus either merge as-is or add the missing constants. And thanks for checking again on PowerPC 970 in 64-bit mode. Turns out I was wrong and it actually does have its own subtype, even in a 64-bit Mach-O file. :+1:

woodruffw:
Great! I'll fill this PR in a bit more and merge it in a bit.

PR #28, opened 2016-06-17
test: Add missing '*_expected*' target and binaries

UniqMartin:
Add the currently missing target to the `Makefile`, generate the matching binaries (by manually invoking `install_name_tool` as a recompile with the new target would also change the binary used for comparison, due to different compilers and other environmental conditions) to resemble the already existing such file for `x86_64`, and use them in the placeholder test `test_change_rpath`.

cc @woodruffw


woodruffw:
Looks great to me! Going to squash-merge this presently.

UniqMartin:
Thanks!

PR #30, opened 2016-06-20
Make endianness an attribute of all LCs.

woodruffw:
This was an oversight on my part, but we're going to need `endianness` in a good number
of load commands in order to extract things like symbol/hint tables in an encapsulated fashion (i.e., not returning flow back to the main `MachOFile`/`FatFile` object every time we need data relevant to a LC).

Once it's in, I can open PRs for my branches that implement the tables/internal structures relevant to various LCs.

cc @UniqMartin

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -148,8 +148,8 @@ module MachO
   # represented, and no actual data. Used when a more specific class
   # isn't available/implemented.
   class LoadCommand < MachOStructure
-    # @return [Fixnum] the offset in the file the command was created from
-    attr_reader :offset
+    # @return [MachO::MachOView] the raw view associated with the load command
+    attr_reader :view
 
     # @return [Fixnum] the load command's identifying number
     attr_reader :cmd
@@ -161,28 +161,32 @@ module MachO
     SIZEOF = 8
 
     # Creates a new LoadCommand given an offset and binary string
-    # @param raw_data [String] the raw Mach-O data
-    # @param endianness [Symbol] the endianness of the command (:big or :little)
-    # @param offset [Fixnum] the offset to initialize with
+    # @param view [MachO::MachOView] the load command's raw view
     # @param bin [String] the binary string to initialize with
     # @return [MachO::LoadCommand] the new load command
     # @api private
-    def self.new_from_bin(raw_data, endianness, offset, bin)
-      format = specialize_format(self::FORMAT, endianness)
+    def self.new_from_bin(view)
+      bin = view.raw_data.slice(view.offset, bytesize)
+      format = specialize_format(self::FORMAT, view.endianness)
 
-      self.new(raw_data, offset, *bin.unpack(format))
+      self.new(view, *bin.unpack(format))
     end
 
-    # @param raw_data [String] the raw Mach-O data
-    # @param offset [Fixnum] the offset to initialize with
+    # @param view [MachO::MachOView] the load command's raw view
     # @param cmd [Fixnum] the load command's identifying number
     # @param cmdsize [Fixnum] the size of the load command in bytes
     # @api private
-    def initialize(raw_data, offset, cmd, cmdsize)
-      @raw_data = raw_data
-      @offset = offset
+    def initialize(view, cmd, cmdsize)
+      @view = view
+      @offset = view.offset
       @cmd = cmd
       @cmdsize = cmdsize
+    end
+
+    # @return [Fixnum] the load command's offset in the source file
+    # @deprecated use {#view} instead
+    def offset
+      view.offset
     end
 
     # @return [Symbol] a symbol representation of the load command's identifying number
@@ -235,8 +239,8 @@ module MachO
     SIZEOF = 24
 
     # @api private
-    def initialize(raw_data, offset, cmd, cmdsize, uuid)
-      super(raw_data, offset, cmd, cmdsize)
+    def initialize(view, cmd, cmdsize, uuid)
+      super(view, cmd, cmdsize)
       @uuid = uuid.unpack("C16") # re-unpack for the actual UUID array
     end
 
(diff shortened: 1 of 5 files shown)


UniqMartin:
I have to admit that I'm not exactly thrilled to add yet another argument to the already pretty long argument lists of all these constructors, not to mention that this is pretty repetitive.

I have the feeling that we should aggregate the data a bit. `raw_data`, `endianness`, and `offset` are always passed together and basically represent a view of the underlying raw data, namely “this load command starts at `offset` in the blob `raw_data` and is `endianness`-endian”. How about creating a new class, e.g. named `MachOView`, that combines these attributes and gets passed instead?

In the long run this could also help with abstracting away the direct reads from and writes to the `raw_data` array, making it easier to only read/write the necessary parts of a file on an on-demand basis.

woodruffw:
> I have to admit that I'm not exactly thrilled to add yet another argument to the already pretty long argument lists of all these constructors, not to mention that this is pretty repetitive.

I agree. 

> How about creating a new class, e.g. named MachOView, that combines these attributes and gets passed instead?

This is what I was thinking as well, but I couldn't think of the word for it :sweat_smile:. `MachOView` is perfect! I'll get to work on this.

lib/macho/macho_file.rb
```diff
@@ -395,7 +395,8 @@ def get_load_commands
         # why do I do this? i don't like declaring constants below
         # classes, and i need them to resolve...
         klass = MachO.const_get "#{LC_STRUCTURES[cmd_sym]}"
-        command = klass.new_from_bin(@raw_data, endianness, offset, @raw_data.slice(offset, klass.bytesize))
+        view = MachOView.new(@raw_data, endianness, offset)
+        command = klass.new_from_bin(view, @raw_data.slice(offset, klass.bytesize))
```
UniqMartin: This looks a lot like the `@raw_data.slice(…)` should be in `LoadCommand#new_from_bin` instead of being passed as a second argument.

lib/macho/view.rb
```diff
@@ -0,0 +1,23 @@
+module MachO
```
UniqMartin: The indentation in this file is weirdly off. Let's stick to two spaces per level of indentation. :wink:

lib/macho/macho_file.rb
woodruffw: Yep, fixing that now.

lib/macho/view.rb
woodruffw: Yep, tabs. My editor seems to have forgotten my preferences for this repo...

lib/macho/load_commands.rb
```diff
@@ -148,7 +148,11 @@ module MachO
   # represented, and no actual data. Used when a more specific class
   # isn't available/implemented.
   class LoadCommand < MachOStructure
-    # @return [Fixnum] the offset in the file the command was created from
+    # @return [MachO::MachOView] the raw view associated with the load command
+    attr_reader :view
+
+    # @return [Fixnum] the load command's offset in the source file
+    # @deprecated use {#view} instead
     attr_reader :offset
```
UniqMartin: Instead of duplicating the data, create a method `offset` that transparently forwards the query to the view, e.g.:

```ruby
def offset
  view.offset
end
```

Or if you want to fully embrace Ruby's standard library, have a look at the `Forwardable` module.

lib/macho/view.rb
UniqMartin: http://editorconfig.org/ and an `.editorconfig` file in the repository root to the rescue?

woodruffw: This looks great! I'll open a PR for this as well.

lib/macho/load_commands.rb
woodruffw: I have somehow managed to never use `Forwardable`, but it looks like it could make a lot of `ruby-macho`'s delegation a lot simpler. I'm going to make a note of it, but go with the `def` until I know it well enough to roll it out everywhere :smile:.

UniqMartin: `Forwardable` might be problematic for documentation, but otherwise it's really easy to use.

woodruffw: Hmm, yeah. Making the documentation more indirect/complex might be a dealbreaker.

UniqMartin:
This generally looks good to me, but I'll do a proper review a bit later. Thus please don't merge just yet.

UniqMartin:
I'm happy with the code structure, i.e. in this regard the PR is ready to merge. Would be nice if you could review the method signatures/calls one more time and rewrap the argument lists where appropriate. The argument lists became shorter and it feels silly to wrap them across multiple lines in some instances.

> Once it's in, I can open PRs for my branches that implement the tables/internal structures relevant to various LCs.

If you already have something ready (or almost ready), can you provide a sneak peek of how this will look like? Would be interesting to see how the things provided in this PR will be used.

woodruffw:
> Would be nice if you could review the method signatures/calls one more time and rewrap the argument lists where appropriate. The argument lists became shorter and it feels silly to wrap them across multiple lines in some instances.

Sure thing!

> If you already have something ready (or almost ready), can you provide a sneak peek of how this will look like?

Certainly. I'll push a branch with changes tomorrow, but the structure is inspired by the nesting of `LCStr` (as `MachO::LoadCommand::LCStr`). For example, `TwolevelHintsCommand` will have a `TwolevelHintsCommand::TwolevelHintsTable`, which will then encapsulate an array of `TwolevelHints`. That's pretty verbose, so I may abbreviate/truncate to keep things brief, but the general structure is there.

UniqMartin:
👏 — Don't forget to [clean up your remote branches](https://github.com/Homebrew/ruby-macho/branches) once things have been successfully merged.

woodruffw:
Cleaned up, thanks for reminding me!

PR #31, opened 2016-06-20
Add .editorconfig.

woodruffw:
I mostly copied from the EditorConfig [example](http://editorconfig.org/#example-file), so tell me if I missed anything :smile:.

cc @UniqMartin

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.rb]
+charset = utf-8
+indent_style = space
+indent_size = 2
+
+[Makefile]
+indent_style = tab


.editorconfig
```diff
@@ -0,0 +1,13 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
```
UniqMartin: I would additionally add `trim_trailing_whitespace = true` here.

UniqMartin:
One addition, otherwise I'm happy with the adapted example. 😸

Just out of curiosity, what are you actually using as your programmer's editor of choice?

woodruffw:
I currently use Sublime Text 3 for most of my development, and I occasionally switch back to my original editor (gvim). Most of `ruby-macho` was written in ST, though :smile:.

UniqMartin:
Nice! I'm also predominantly a Sublime Text user. :+1:

PR #32, opened 2016-06-21
Add direct two-level hints table access.

woodruffw:
This one is *probably* safe to merge, but I'd like some tests before I do so.
Do you have any ideas on getting `clang`/`ld` to emit a two-level namespace hints lookup table?

cc @UniqMartin

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -167,7 +167,7 @@ module MachO
     # @api private
     def self.new_from_bin(view)
       bin = view.raw_data.slice(view.offset, bytesize)
-      format = specialize_format(self::FORMAT, view.endianness)
+      format = Utils.specialize_format(self::FORMAT, view.endianness)
 
       self.new(view, *bin.unpack(format))
     end
@@ -758,9 +758,9 @@ module MachO
       # @param nhints [Fixnum] the number of two-level hints in the table
       # @api private
       def initialize(view, htoffset, nhints)
-        format = (view.endianness == :little) ? "V<" : "V>"
-        raw_table = view.raw_data.slice((htoffset...(htoffset + (nhints * 4))))
-        blobs = raw_table.unpack(format * nhints)
+        format = Utils.specialize_format("L=#{nhints}", view.endianness)
+        raw_table = view.raw_data[htoffset, nhints * 4]
+        blobs = raw_table.unpack(format)
 
         @hints = blobs.map { |b| TwolevelHint.new(b) }
       end
@@ -776,8 +776,8 @@ module MachO
         # @param blob [Fixnum] the 32-bit number containing the lookup hint
         # @api private
         def initialize(blob)
-          @isub_image = (blob >> 24)
-          @itoc = (blob & ((1 << 24) - 1))
+          @isub_image = blob >> 24
+          @itoc = blob & 0x00FFFFFF
         end
       end
     end
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -733,6 +733,9 @@ module MachO
     # @return [Fixnum] the number of hints in the hint table
     attr_reader :nhints
 
+    # @return [MachO::TwolevelHintsCommand::TwolevelHintTable] the hint table
+    attr_reader :table
+
     FORMAT = "L=4"
     SIZEOF = 16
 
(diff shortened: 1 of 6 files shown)


lib/macho/load_commands.rb
```diff
@@ -741,6 +744,42 @@ def initialize(view, cmd, cmdsize, htoffset, nhints)
...
+    class TwolevelHintsTable
+      # @return [Array<MachO::TwoLevelHintsTable::TwoLevelHint>] all hints in the table
+      attr_reader :hints
+
+      # @param view [MachO::MachOView] the view into the current Mach-O
+      # @param htoffset [Fixnum] the offset of the hints table
+      # @param nhints [Fixnum] the number of two-level hints in the table
+      # @api private
+      def initialize(view, htoffset, nhints)
+        format = (view.endianness == :little) ? "V<" : "V>"
```
UniqMartin: Can we re-use `MachOStructure.specialize_format`, so that we can avoid specifying both little- and big-endian cases over and over again? Keeping this logic mostly in a single location will also make it easier to restore Ruby 1.8.7 compatibility, if we have to (though I certainly hope we can avoid that). Maybe it's worth moving `specialize_format` somewhere else to make it more generally available …

lib/macho/load_commands.rb
```diff
@@ -741,6 +744,42 @@ def initialize(view, cmd, cmdsize, htoffset, nhints)
...
+        raw_table = view.raw_data.slice((htoffset...(htoffset + (nhints * 4))))
```
UniqMartin: In this case, the range makes things worse and the `[start, length]` variant is much more readable, i.e.:

```ruby
raw_table = view.raw_data[htoffset, nhints * 4]
```

(I also personally prefer `[]` over `slice`, but that's less of an issue.)

lib/macho/load_commands.rb
```diff
@@ -741,6 +744,42 @@ def initialize(view, cmd, cmdsize, htoffset, nhints)
...
+        blobs = raw_table.unpack(format * nhints)
```
UniqMartin: You definitely don't want to create a very long format string here. If you wanted to be very explicit, then `"#{format}#{nhints}"` will do, but I'd personally prefer just `"#{format}*"` and actually include the `*` in the format string from the very beginning, i.e. `"L=*"`.

lib/macho/load_commands.rb
UniqMartin: Oh and you probably meant to use `L<` and `L>` here. `V` is already tied to a specific endianness.

lib/macho/load_commands.rb
```diff
@@ -741,6 +744,42 @@ def initialize(view, cmd, cmdsize, htoffset, nhints)
...
+        attr_reader :isub_image
+
+        # @return [Fixnum] the index into the table of contents
+        attr_reader :itoc
+
+        # @param blob [Fixnum] the 32-bit number containing the lookup hint
+        # @api private
+        def initialize(blob)
+          @isub_image = (blob >> 24)
+          @itoc = (blob & ((1 << 24) - 1))
```
UniqMartin: I think you can safely skip the outer parentheses without making things less readable. `&` is as much a regular binary operator as e.g. `+`. Personally, if the bitmask is known in advance, I'd put it here explicitly instead of “computing” it, i.e. `@itoc = blob & 0x00FFFFFF`. (The leading zeros also make it clear that you're expecting 32-bit input.)

UniqMartin:
> This one is *probably* safe to merge, but I'd like some tests before I do so.

I'd say “not quite” as you can infer from my code comments, as not all of them are purely stylistic.

> Do you have any ideas on getting `clang`/`ld` to emit a two-level namespace hints lookup table?

I'd have to check what older versions are doing and whether I can dig up an ancient binary that uses this load command, but the linker bundled with Xcode 7.3.1 simply ignores the `-twolevel_namespace_hints` flag and the dynamic linker (`dyld`) on OS X 10.11 equally ignores the `LC_TWOLEVEL_HINTS` load command.

lib/macho/load_commands.rb
woodruffw: > Maybe it's worth moving specialize_format somewhere else to make it more generally available

Yeah, this would be best. I've been dropping utility functions in *utils.rb* (...without a module scope, which should be fixed), so I can move `specialize_format` there as well.

> Oh and you probably meant to use L< and L> here. V is already tied to a specific endianness.

Yep, silly mistake.

lib/macho/load_commands.rb
woodruffw: Agreed. I occasionally forget about `String#[]` :sweat_smile:

lib/macho/load_commands.rb
woodruffw: Got it! The parens and `((1 << 24) - 1)` were personal style choices, but I don't feel strongly about them.

lib/macho/load_commands.rb
woodruffw: On second thought, making `TwolevelHintsTable` a subclass of `MachOStructure` shouldn't require too many changes. I'll go with that for now (and table moving `specialize_format` elsewhere).

lib/macho/load_commands.rb
UniqMartin: I don't feel very strongly about this either, but maybe my arguments were persuasive enough. :wink:

lib/macho/load_commands.rb
woodruffw: I do personally prefer being explicit with `#{nhints}` in this case, as it goes generally with the `FORMAT` style we've used so far (i.e., including numbers even when there's only one type being unpacked).

woodruffw:
> I'd say “not quite” as you can infer from my code comments, as not all of them are purely stylistic.

Definitely spoke too soon :smile: 

> I'd have to check what older versions are doing and whether I can dig up an ancient binary that uses this load command, but the linker bundled with Xcode 7.3.1 simply ignores the -twolevel_namespace_hints flag and the dynamic linker (dyld) on OS X 10.11 equally ignores the LC_TWOLEVEL_HINTS load command.

Yep, this is what I gathered while (trying to) make tests.

lib/macho/load_commands.rb
UniqMartin: So far we've only unpacked structures of fixed length, so I'm not sure if this is a good enough argument. Being explicit doesn't buy us much as far as I can tell, as you're being rather explicit in the preceding line when slicing the relevant bytes from the raw data blob and its not like Ruby is going to complain that it could not extract as many elements as you requested (though it will pad the output with `nil` values).

Just some more thoughts. I don't mind with you being explicit here.

woodruffw: > So far we've only unpacked structures of fixed length, so I'm not sure if this is a good enough argument.

Yeah, it falls apart when you consider that.

One other possible argument is that we may end up (not here, but in sections/blobs referenced by `dyld` commands) unpacking non-fixed structures whose fields vary in width.

In a case like that, using `*` would cause the unpacker to be excessively greedy and swallow up fields with incorrect widths (for example, `Q=*L=*`). Being explicit about expected elements here sets a good style precedent for more complex cases like the above.

UniqMartin: You've convinced me. :+1: on being explicit.

PR #33, opened 2016-06-26
Add obsolete and internal load commands.

woodruffw:
Adds `SymsegCommand` (`LC_SYMSEG`), `FvmlimCommand` (`LC_LOADFVMLIB`, `LC_IDFVMLIB`), `IdentCommand` (`LC_IDENT`), and `FvmfileCommand` (`LC_FCMFILE`).

`libmacho` doesn't seem to expose a pubic definition for the structure associated with `LC_PREPAGE`, so that one will have to remain unimplemented for now.

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -72,14 +72,14 @@ module MachO
   LC_STRUCTURES = {
     :LC_SEGMENT => "SegmentCommand",
     :LC_SYMTAB => "SymtabCommand",
-    :LC_SYMSEG => "LoadCommand", # obsolete
+    :LC_SYMSEG => "SymsegCommand", # obsolete
     :LC_THREAD => "ThreadCommand", # seems obsolete, but not documented as such
     :LC_UNIXTHREAD => "ThreadCommand",
-    :LC_LOADFVMLIB => "LoadCommand", # obsolete
-    :LC_IDFVMLIB => "LoadCommand", # obsolete
-    :LC_IDENT => "LoadCommand", # obsolete
-    :LC_FVMFILE => "LoadCommand", # reserved for internal use only
-    :LC_PREPAGE => "LoadCommand", # reserved for internal use only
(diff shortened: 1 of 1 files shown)


lib/macho/load_commands.rb
```diff
@@ -1008,4 +1008,68 @@ def version_string
...
+    # @api private
+    def initialize(view, cmd, cmdsize, offset, size)
+      super(view, cmd, cmdsize)
+      @offset = offset
+      @size = size
+    end
+  end
+
+  # An obsolete load command containing a free format string table. Each string
+  # is null-terminated and the command is null-aligned to a multiple of 4.
```
UniqMartin: I think “zero-padded” is clearer than “null-aligned” or am I misunderstanding this? Any reason why you omitted the “Corresponds to `LC_*`.” comment here and in the next two commands?

UniqMartin:
:+1: Feel free to merge once the minor documentation-related comment has been addressed.

lib/macho/load_commands.rb
woodruffw: > I think “zero-padded” is clearer than “null-aligned” or am I misunderstanding this?

I agree, I'll change this.

The lack of the "corresponds" lines was an oversight as well, I'll add them.

lib/macho/load_commands.rb
```diff
@@ -1046,4 +1046,73 @@ def version_string
...
+    # @return [Fixnum] the library's header address
+    attr_reader :header_addr
+
+    FORMAT = "L=5"
+    SIZEOF = 20
+
+    def initialize(view, cmd, cmdsize, name, minor_version, header_addr)
+      super(view, cmd, cmdsize)
+      @name = LCStr.new(self, name)
+      @header_addr = header_addr
```
UniqMartin: Dang, I noticed this just now, but what happened to `minor_version`?

woodruffw: Eek. Fixed it in bcfcb0e.

PR #34, opened 2016-06-28
LCStr: Refactor and simplify

UniqMartin:
- Drop constructor argument `raw_data` as it is redundant and the data can be obtained from the load command via `lc.view.raw_data` (and adapt related load commands accordingly).
- Use intermediate variables to shorten overly long line.
- Avoid storing constructor arguments in instance variables if they are not going to be referenced later (`@raw_data` and `@lc`).
- Improve names of instance variables.

@woodruffw Sorry that this is somewhat interfering with #33 where you add new load commands that also use `LCStr` (but obviously still using the three-argument constructor).

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -204,25 +204,25 @@ module MachO
     # pretend that strings stored in LCs are immediately available without
     # explicit operations on the raw Mach-O data.
     class LCStr
-      # @param raw_data [String] the raw Mach-O data.
       # @param lc [MachO::LoadCommand] the load command
       # @param lc_str [Fixnum] the offset to the beginning of the string
       # @api private
-      def initialize(raw_data, lc, lc_str)
-        @raw_data = raw_data
-        @lc = lc
-        @lc_str = lc_str
-        @str = @raw_data.slice(@lc.offset + @lc_str...@lc.offset + @lc.cmdsize).delete("\x00")
+      def initialize(lc, lc_str)
+        view = lc.view
+        lc_str_abs = view.offset + lc_str
+        lc_end = view.offset + lc.cmdsize - 1
+        @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
+        @string_offset = lc_str
       end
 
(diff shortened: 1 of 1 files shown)


woodruffw:
> Sorry that this is somewhat interfering with #33 where you add new load commands that also use LCStr (but obviously still using the three-argument constructor).

Not a problem! I'll merge this first and update #33 accordingly :+1:

lib/macho/load_commands.rb
```diff
@@ -204,25 +204,25 @@ def to_s
...
       # @api private
-      def initialize(raw_data, lc, lc_str)
-        @raw_data = raw_data
-        @lc = lc
-        @lc_str = lc_str
-        @str = @raw_data.slice(@lc.offset + @lc_str...@lc.offset + @lc.cmdsize).delete("\x00")
+      def initialize(lc, lc_str)
+        raw_data = lc.view.raw_data
+        lc_str_abs = lc.offset + lc_str
+        lc_end = lc.offset + lc.cmdsize - 1
```
UniqMartin: Upon closer inspection, I wonder if this line and the line above this one should directly use `lc.view.offset` instead of relying on the deprecated `LoadCommand#offset` method.

woodruffw: Yeah, I think that would be preferable.

UniqMartin:
Updated! Merge at your convenience once you're happy with this PR.

woodruffw:
:+1:

UniqMartin:
Thanks! Keep in mind that you can switch between a regular merge and a squash merge when merging a PR. Since this was only a single commit, a squash merge would have slightly simplified the history without any real loss of information. (A regular merge that preserves individual commits sometimes also makes sense, particularly in a more complex PR with individual commits building on top of each other.)

woodruffw:
Whoops. I'll remember that for future merges. I keep flipping it back to regular merge as a force of habit...

UniqMartin:
No worries! It's way better like this that doing the reverse and accidentally squashing a carefully crafted multi-commit history into a single commit, so all context is lost.

PR #37, opened 2016-07-01
macho_file: Add low_fileoff method, use String#bytesize for LC strings.

woodruffw:
Some more triage on `set_lc_str_in_cmd` while I work on an ultimate replacement.

Having a discrete `MachOFIle#low_fileoff` method will probably come in handy, since `low_fileoff` appears in a few places in `libmacho` and will certainly come in handy for the upcoming `MachOFile#add_command` (or whatever we settle on).

Replacing `String#size` with `String#bytesize` doesn't cause any of our current tests to break, and is probably the best temporary fix for any internationalization issues we come across.

cc @UniqMartin

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -404,6 +404,26 @@ module MachO
       load_commands
     end
 
+    # The low file offset (offset to first section data).
+    # @return [Fixnum] the offset
+    # @private
+    def low_fileoff
+      offset = @raw_data.size
+
+      segments.each do |seg|
+        sections(seg).each do |sect|
+          next if sect.size == 0
+          next if sect.flag?(:S_ZEROFILL)
+          next if sect.flag?(:S_THREAD_LOCAL_ZEROFILL)
+          next unless sect.offset < offset
+
+          offset = sect.offset
+        end
+      end
+
+      offset
+    end
+
     # Updates the size of all load commands in the raw data.
     # @param size [Fixnum] the new size, in bytes
     # @return [void]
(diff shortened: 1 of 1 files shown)


lib/macho/macho_file.rb
```diff
@@ -497,13 +504,13 @@ def set_lc_str_in_cmd(cmd, lc_str, old_str, new_str)
       @raw_data[cmd.offset + 4, 4] = [new_padded_size].pack(fmt)
 
       # delete the old str
-      @raw_data.slice!(cmd.offset + lc_str.to_i...cmd.offset + cmd.class.bytesize + old_str.size)
+      @raw_data.slice!(cmd.offset + lc_str.to_i...cmd.offset + cmd.class.bytesize + old_str.bytesize)
```
UniqMartin: While you're here, would you mind fixing the inconsistency that you're using `lc_str.to_i` on the left-hand side and `lc_str.class.bytesize` on the right-hand side? (I think both should be `lc_str.to_i` and just accidentally coincide in the cases we've come across so far.)

Better yet, use a different signature of `String#slice!` to avoid some repetition:

```ruby
@raw_data.slice!(cmd.offset + lc_str.to_i, old_str.bytesize)
```

UniqMartin:
Commented one minor inconsistency that can be optimized away, but otherwise: :+1:

lib/macho/macho_file.rb
woodruffw: Fixed!

UniqMartin:
:tada:

PR #38, opened 2016-07-03
Add LoadCommand.create for creating independent load commands.

woodruffw:
- [x] Creation of load commands without a view.
- [x] Serialization of created load commands.

Downsides to this approach:
* Argument order is currently unchecked, which means that users will almost certainly run into confusing bugs if they provide arguments in the order not used by structures in `libmacho` (and `ruby-macho`).

cc @UniqMartin

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -57,6 +57,8 @@ module MachO
     0x30 => :LC_VERSION_MIN_WATCHOS,
   }.freeze
 
+  LOAD_COMMAND_CONSTANTS = LOAD_COMMANDS.invert.freeze
+
   # load commands responsible for loading dylibs
   # @api private
   DYLIB_LOAD_COMMANDS = [
@@ -65,6 +67,14 @@ module MachO
     :LC_REEXPORT_DYLIB,
     :LC_LAZY_LOAD_DYLIB,
     :LC_LOAD_UPWARD_DYLIB,
+  ].freeze
+
+  # load commands that can be created manually via {LoadCommand.create}
+  # @api private
+  CREATABLE_LOAD_COMMANDS = DYLIB_LOAD_COMMANDS + [
+    :LC_ID_DYLIB,
+    :LC_RPATH,
+    :LC_LOAD_DYLINKER,
   ].freeze
 
   # association of load command symbols to string representations of classes
@@ -160,7 +170,7 @@ module MachO
     FORMAT = "L=2"
     SIZEOF = 8
 
-    # Creates a new LoadCommand given an offset and binary string
+    # Instantiates a new LoadCommand given a view into its origin Mach-O
     # @param view [MachO::MachOView] the load command's raw view
     # @return [MachO::LoadCommand] the new load command
     # @api private
@@ -171,6 +181,23 @@ module MachO
       self.new(view, *bin.unpack(format))
     end
 
+    # Creates a new (viewless) command corresponding to the symbol provided
+    # @param cmd_sym [Symbol] the symbol of the load command being created
+    # @param args [Array] the arguments for the load command being created
+    def self.create(cmd_sym, *args)
+      raise LoadCommandNotCreatableError.new(cmd_sym) unless CREATABLE_LOAD_COMMANDS.include?(cmd_sym)
+
+      klass = MachO.const_get "#{LC_STRUCTURES[cmd_sym]}"
+      cmd = LOAD_COMMAND_CONSTANTS[cmd_sym]
+
+      # cmd will be filled in, view and cmdsize will be left unpopulated
+      klass_arity = klass.instance_method(:initialize).arity - 3
+
+      raise LoadCommandCreationArityError.new(cmd_sym, klass_arity, args.size) if klass_arity != args.size
+
+      klass.new(nil, cmd, nil, *args)
+    end
+
     # @param view [MachO::MachOView] the load command's raw view
     # @param cmd [Fixnum] the load command's identifying number
     # @param cmdsize [Fixnum] the size of the load command in bytes
@@ -179,6 +206,22 @@ module MachO
       @view = view
       @cmd = cmd
       @cmdsize = cmdsize
+    end
+
+    # @return [Boolean] true if the load command can be serialized, false otherwise
+    def serializable?
+      CREATABLE_LOAD_COMMANDS.include?(LOAD_COMMANDS[cmd])
+    end
+
+    # @param context [MachO::LoadCommand::SerializationContext] the context
+    #  to serialize into
+    # @return [String, nil] the serialized fields of the load command, or nil
+    #  if the load command can't be serialized
+    # @api private
+    def serialize(context)
+      raise LoadCommandNotSerializableError.new(LOAD_COMMANDS[cmd]) unless serializable?
+      format = Utils.specialize_format(FORMAT, context.endianness)
+      [cmd, SIZEOF].pack(format)
     end
 
     # @return [Fixnum] the load command's offset in the source file
@@ -205,14 +248,24 @@ module MachO
     # explicit operations on the raw Mach-O data.
     class LCStr
       # @param lc [MachO::LoadCommand] the load command
-      # @param lc_str [Fixnum] the offset to the beginning of the string
+      # @param lc_str [Fixnum, String] the offset to the beginning of the string,
+      #  or the string itself if not being initialized with a view.
+      # @todo devise a solution such that the `lc_str` parameter is not
+      #  interpreted differently depending on `lc.view`. The current behavior
+      #  is a hack to allow viewless load command creation.
       # @api private
       def initialize(lc, lc_str)
         view = lc.view
-        lc_str_abs = view.offset + lc_str
-        lc_end = view.offset + lc.cmdsize - 1
-        @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
-        @string_offset = lc_str
+
+        if view
+          lc_str_abs = view.offset + lc_str
+          lc_end = view.offset + lc.cmdsize - 1
+          @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
+          @string_offset = lc_str
+        else
+          @string = lc_str
+          @string_offset = 0
+        end
       end
 
       # @return [String] a string representation of the LCStr
@@ -223,6 +276,30 @@ module MachO
       # @return [Fixnum] the offset to the beginning of the string in the load command
       def to_i
         @string_offset
+      end
+    end
+
+    # Represents the contextual information needed by a load command to
+    # serialize itself correctly into a binary string.
+    class SerializationContext
+      # @return [Symbol] the endianness of the serialized load command
+      attr_reader :endianness
+
+      # @return [Fixnum] the constant alignment value used to pad the serialized load command
+      attr_reader :alignment
+
+      # @param macho [MachO::MachOFile] the file to contextualize
+      # @return [MachO::LoadCommand::SerializationContext] the resulting context
+      def self.context_for(macho)
+        self.new(macho.endianness, macho.alignment)
+      end
+
+      # @param endianness [Symbol] the endianness of the context
+      # @param alignment [Fixnum] the alignment of the context
+      # @api private
+      def initialize(endianness, alignment)
+        @endianness = endianness
+        @alignment = alignment
       end
     end
   end
@@ -399,6 +476,17 @@ module MachO
       @current_version = current_version
       @compatibility_version = compatibility_version
     end
+
+    # @param context [MachO::LoadCcommand::SerializationContext] the context
+    # @return [String] the serialized fields of the load command
+    # @api private
+    def serialize(context)
+      format = Utils.specialize_format(FORMAT, context.endianness)
+      string_payload, string_offsets = Utils.pack_strings(SIZEOF, context.alignment, :name => name.to_s)
+      cmdsize = SIZEOF + string_payload.bytesize
+      [cmd, cmdsize, string_offsets[:name], timestamp, current_version,
+        compatibility_version].pack(format) + string_payload
+    end
   end
 
   # A load command representing some aspect of the dynamic linker, depending
@@ -415,6 +503,16 @@ module MachO
     def initialize(view, cmd, cmdsize, name)
       super(view, cmd, cmdsize)
       @name = LCStr.new(self, name)
+    end
+
+    # @param context [MachO::LoadCcommand::SerializationContext] the context
+    # @return [String] the serialized fields of the load command
+    # @api private
+    def serialize(context)
+      format = Utils.specialize_format(FORMAT, context.endianness)
+      string_payload, string_offsets = Utils.pack_strings(SIZEOF, context.alignment, :name => name.to_s)
+      cmdsize = SIZEOF + string_payload.bytesize
+      [cmd, cmdsize, string_offsets[:name]].pack(format) + string_payload
     end
   end
 
@@ -811,6 +909,16 @@ module MachO
     def initialize(view, cmd, cmdsize, path)
       super(view, cmd, cmdsize)
       @path = LCStr.new(self, path)
+    end
+
+    # @param context [MachO::LoadCcommand::SerializationContext] the context
+    # @return [String] the serialized fields of the load command
+    # @api private
+    def serialize(context)
+      format = Utils.specialize_format(FORMAT, context.endianness)
+      string_payload, string_offsets = Utils.pack_strings(SIZEOF, context.alignment, :path => path.to_s)
+      cmdsize = SIZEOF + string_payload.bytesize
+      [cmd, cmdsize, string_offsets[:path]].pack(format) + string_payload
     end
   end
 
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -64,6 +64,10 @@ module MachO
       Utils.magic64?(header.magic)
     end
 
+    def alignment
+      magic32? ? 4 : 8
+    end
+
     # @return [Boolean] true if the file is of type `MH_OBJECT`, false otherwise
     def object?
       header.filetype == MH_OBJECT
@@ -464,12 +468,6 @@ module MachO
     # @todo This needs to be replaced.
     # @see https://github.com/Homebrew/ruby-macho/pull/35
     def set_lc_str_in_cmd(cmd, lc_str, old_str, new_str)
-      if magic32?
-        cmd_round = 4
-      else
-        cmd_round = 8
-      end
-
       new_sizeofcmds = header.sizeofcmds
       old_str = old_str.dup
       new_str = new_str.dup
@@ -478,8 +476,8 @@ module MachO
       new_prepad = cmd.class.bytesize + new_str.bytesize + 1
 
       # calculate the original and new padded sizes of the strings
-      old_padded_size = Utils.round(old_prepad, cmd_round)
-      new_padded_size = Utils.round(new_prepad, cmd_round)
+      old_padded_size = Utils.round(old_prepad, alignment)
+      new_padded_size = Utils.round(new_prepad, alignment)
 
       # calculate the number of pad bytes used in the old and new strings
       old_pad = old_padded_size - (old_prepad)
--- a/lib/macho/utils.rb
+++ b/lib/macho/utils.rb
@@ -11,6 +11,13 @@ module MachO
       value
     end
 
+    # @param size [Fixnum] the unpadded size
+    # @param alignment [Fixnum] the number to alignment the size with
+    # @return [Fixnum] the number of pad bytes required
+    def self.padding_for(size, alignment)
+      round(size, alignment) - size
+    end
+
     # Convert an abstract (native-endian) String#unpack format to big or little.
     # @param format [String] the format string being converted
     # @param endianness [Symbol] either `:big` or `:little`
@@ -18,6 +25,26 @@ module MachO
     def self.specialize_format(format, endianness)
       modifier = (endianness == :big) ? ">" : "<"
       format.tr("=", modifier)
+    end
+
+    # @param fixed_offset [Fixnum] the baseline offset for the first packed string
+    # @param alignment [Fixnum] the alignment value to use for packing
+    # @param strings [Hash] the labeled strings to pack
+    # @return Array<String, Hash> the packed string and labeled offsets
+    def self.pack_strings(fixed_offset, alignment, strings = {})
+      offsets = {}
+      next_offset = fixed_offset
+      payload = ""
+
+      strings.each do |key, string|
+        offsets[key] = next_offset
+        payload << string
+        payload << "\x00"
+        next_offset += string.bytesize + 1
+      end
+
+      payload << "\x00" * padding_for(fixed_offset + payload.bytesize, alignment)
+      [payload, offsets]
     end
 
     # @param num [Fixnum] the number being checked
--- a/test/test_serialize_load_commands.rb
+++ b/test/test_serialize_load_commands.rb
@@ -1,0 +1,257 @@
+require "minitest/autorun"
+require "helpers"
+require "macho"
+
+class MachOLoadCommandSerializationTest < Minitest::Test
+  include Helpers
+
+  def test_can_serialize
+    filename = fixture(:i386, "hello.bin")
+    file = MachO::MachOFile.new(filename)
+    lc = file[:LC_SEGMENT].first
+
+    refute lc.serializable?
+
+    assert_raises MachO::LoadCommandNotSerializableError do
+      lc.serialize(MachO::LoadCommand::SerializationContext.context_for(file))
+    end
+  end
+
+  def test_serialize_segment
+    pass
+  end
+
+  def test_serialize_symtab
+    pass
+  end
+
+  def test_serialize_symseg
+    pass
+  end
+
+  def test_serialize_thread
+    pass
+  end
+
+  def test_serialize_unixthread
+    pass
+  end
+
+  def test_serialize_loadfvmlib
+    pass
+  end
+
+  def test_serialize_ident
+    pass
+  end
+
+  def test_serialize_fvmfile
+    pass
+  end
+
+  def test_serialize_prepage
+    pass
+  end
+
+  def test_serialize_dysymtab
+    pass
+  end
+
+  def test_serialize_load_dylib
+    filenames = SINGLE_ARCHES.map { |a| fixture(a, "hello.bin") }
+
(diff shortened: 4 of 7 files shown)


woodruffw:
This approach to serialization is fairly verbose, but the other alternatives I considered mostly involved a lot of fragile metaprogramming (e.g., inspecting each load command's `#instance_variables` and using those with `#send`). This could definitely be trimmed down a bit, and I'll do that when I'm slightly less sleep deprived :smile:

Serialization of load commands created with `LoadCommand#create` does not currently work, as they're not padded correctly (i.e., the `npad` calculation in `LCStr#initialize` is only correct when `lc.cmdsize` is initially correct and it isn't). I think the best fix for this would probably be an introduction of `cmd_round` somewhere (a la `MachOFile#set_lc_str_in_cmd`), possibly in a subclass of `MachOView` that we then differentiate in some way from a "real" view in order to preserve the current logic.

woodruffw:
An equally good (or perhaps better, since it better matches the values for "real" LCs) fix for `lc.cmdsize` in created load commands might be to simply pass `cmd_round` into `LoadCommand.create` as an argument.

UniqMartin:
The implementation you pushed is already close to what I was hoping to arrive at. :+1: I'm also happy with the general level of verboseness for now, though some simplification may be possible in the future once some more infrastructure is in place (the mini-DSL mentioned in https://github.com/Homebrew/ruby-macho/pull/24#issuecomment-223783685 could help with this).

I'll go ahead and provide more thorough feedback in code comments and then finish off with a few more remarks in a concluding comment. *(If you happen to read this before I made that concluding comment, I think it makes sense to wait with responding to any of my comments until you have the full picture.)*

woodruffw:
> though some simplification may be possible in the future once some more infrastructure is in place (the mini-DSL mentioned in #24 (comment) could help with this).

Yep! This is ultimately what I want the load command API to look like, although I haven't yet decided an approach that results in the least breakage (possibly an overdue 1.0 release?)

lib/macho/load_commands.rb
```diff
@@ -171,6 +27,33 @@ def self.new_from_bin(view)
...
+    # Creates a new (viewless) command corresponding to the symbol provided
+    # @param cmd_sym [Symbol] the symbol of the load command being created
+    # @param args [Array] the arguments for the load command being created
+    # @note this does not currently work. DO NOT USE IT.
+    def self.create(cmd_sym, *args)
+      raise LoadCommandNotCreatableError.new(cmd_sym) unless CREATABLE_LOAD_COMMANDS.include?(cmd_sym)
+
+      klass = MachO.const_get "#{LC_STRUCTURES[cmd_sym]}"
+      cmd = LOAD_COMMAND_CONSTANTS[cmd_sym]
+      cmdsize = klass.bytesize
```
UniqMartin: I don't think it's wise to set up `cmdsize` in this method. Let's just make it `nil` and delay figuring out `cmdsize` until the load command created here gets serialized. That's the earliest we can know the size, as it can be influenced (among other things) by the required padding that depends on whether we're serializing for a 32-bit or 64-bit Mach-O binary.

lib/macho/load_commands.rb
```diff
@@ -171,6 +27,33 @@ def self.new_from_bin(view)
...
+
+      # this is a hack, since the only currently creatable load commands are
+      # ones that contain strings.
+      args.each do |arg|
+        if arg.is_a?(String)
+          cmdsize += arg.bytesize
+        end
+      end
```
UniqMartin: Let's drop this hack. This belongs into the serialization, where we'll have enough information to compute this properly.

lib/macho/load_commands.rb
```diff
@@ -204,15 +97,30 @@ def to_s
     # pretend that strings stored in LCs are immediately available without
     # explicit operations on the raw Mach-O data.
     class LCStr
+      # @return [Fixnum] the number of null bytes used to pad the string
+      attr_reader :npad
```
UniqMartin: Similarly to `cmdsize`, I don't think the string needs to know its own padding. That's assuming too much (among other things that there's only exactly one `LCStr` in any given load command).

lib/macho/load_commands.rb
```diff
@@ -399,6 +339,18 @@ def initialize(view, cmd, cmdsize, name, timestamp, current_version, compatibili
       @current_version = current_version
       @compatibility_version = compatibility_version
     end
+
+    # @param endianness [Symbol, nil] the endianness of the serialized string,
+    #  or nil to use the view's endianness
+    # @return [String] the serialized fields of the load command
+    # @api private
+    def serialize(endianness = nil)
+      endianness ||= view.endianness
```
UniqMartin: *I'm commenting on this specific `serialize` method, but of course everything also applies to all the other `serialize` methods.*

I think we need to pass something that knows about the endianness and bit width of the Mach-O file this load command will be inserted into. To me, the easiest solution seems to be to just pass the `MachOFile` instance for querying this information (though we might want to abstract this away in the future). I also think we shouldn't even try to query the `view` for this information. It won't typically exist.

lib/macho/load_commands.rb
```diff
@@ -399,6 +339,18 @@ def initialize(view, cmd, cmdsize, name, timestamp, current_version, compatibili
...
+      format = Utils.specialize_format(FORMAT, endianness)
+      padded_name = name.to_s + "\x00" * name.npad
+      [cmd, cmdsize, name.to_i, timestamp, current_version,
+        compatibility_version].pack(format) + padded_name
```
UniqMartin: One problem this currently has is that `name.to_i` (`name#@string_offset`) is currently `0` and isn't updated, another is that the padding isn't adjusted to the bit width of the target Mach-O file this will be serialized into.

Here's a suggestion for how this part of the serialization could work (and avoid hard-coding the padding logic into every `serialize` method that has to deal with `LCStr`s):

```ruby
string_payload, string_offsets = Utils.pack_strings(SIZEOF, target.magic32?, :name => name.to_s)
cmdsize = SIZEOF + string_payload.bytesize
[cmd, cmdsize, string_offsets[:name], timestamp, current_version,
  compatibility_version].pack(format) + string_payload
```

This means all the details of computing the offset where a string starts, packing (possibly multiple) strings into a byte stream, then padding this byte stream with zeros to make the load command meet alignment requirements would all be done in the utility function. This avoids encoding too many details into every single `serialize` method. The implementation could look something like (untested):

```ruby
def self.pack_strings(fixed_size, is_32bit, strings = {})
  offsets = {}
  next_offset = fixed_size
  payload = ""
  strings.each do |key, string|
    offsets[key] = next_offset
    payload << string
    payload << "\x00"
    next_offset += string.bytesize + 1
  end
  payload << "\x00" * compute_required_padding(fixed_size + payload.bytesize, is_32bit)
  [payload, offsets]
end
```

lib/macho/load_commands.rb
```diff
@@ -171,6 +27,35 @@ def self.new_from_bin(view)
       self.new(view, *bin.unpack(format))
     end
 
+    # Creates a new (viewless) command corresponding to the symbol provided
+    # @param cmd_sym [Symbol] the symbol of the load command being created
+    # @param args [Array] the arguments for the load command being created
+    # @note this does not currently work. DO NOT USE IT.
+    def self.create(cmd_sym, cmd_round, *args)
```
UniqMartin: This is probably obvious from my previous comments, but I don't think we should hard-code the required alignment on creation. All target-specific encoding should really only happen in the `serialize` method. This will allow creation of a target-agnostic load command that can be then serialized to all Mach-O slices of a given fat binary, e.g. when adding an `LC_RPATH` command to a fat binary. But really the point here is that I'd like the load command to not make any assumptions about how it is going to be serialized until it actually gets serialized.

lib/macho/load_commands.rb
woodruffw: Agreed. My only justification was that `npad` only *needs* to be calculated once per `LCStr` instance, and that's hardly a significant performance concern.

lib/macho/load_commands.rb
woodruffw: > To me, the easiest solution seems to be to just pass the MachOFile instance

I agree that this would be the easiest solution, but I think that it establishes a confusing object relationship - we'd be passing in a whole file instance to create an independent load command whose only relationship with the file is a similarity of intrinsics (endianness, rounding size, etc). 

It would also require us to do something like `lc.serialize(this)` when dealing with created load commands inside a `MachOFile`, and I think that's pretty inelegant compared to `lc.serialize(endianness)`.

test/test_serialize_load_commands.rb
```diff
@@ -0,0 +1,323 @@
...
+  end
+
+  def test_serialize_source_version
+    # ppc does not have a LC_SOURCE_VERSION command
+    filenames = [:i386, :x86_64].map { |a| fixture(a, "hello.bin") }
+
+    filenames.each do |filename|
+      file = MachO::MachOFile.new(filename)
+      lc = file[:LC_SOURCE_VERSION].first
+      blob = lc.view.raw_data[lc.view.offset, lc.cmdsize]
```
UniqMartin: This comes up frequently enough that we might want to add a `raw_data` method to a load command, that uses its `view` and its `cmdsize` attributes to extract the raw bytes representing the load command from the view's raw data.

lib/macho/load_commands.rb
UniqMartin: Performance isn't really my concern here. My concern is that the padding inside a load command has nothing to do with individual strings represented by `LCStr`. The padding is something that is applied to the load command as a whole to make its size a multiple of either 4 or 8. `LCStr` simply isn't the right place to deal with this padding requirement.

lib/macho/load_commands.rb
woodruffw: If we calculate `cmdsize` during serialization instead of creation, load commands created with `LoadCommand.create` never really become complete in their own right as instances. I don't think this is a *huge* problem, but I also think it'll be nice to be able to treat `create`d load commands as first class citizens.

lib/macho/load_commands.rb
woodruffw: Whoops. The comment I left on the next below addresses this.

lib/macho/load_commands.rb
woodruffw: Got it, I'll fix this :+1:

lib/macho/load_commands.rb
UniqMartin: One problem here is that `endianness` alone isn't sufficient to get the proper serialization. To me it doesn't feel that inelegant to tell the `serialize` method something like “here's the file that you will be serialized into; please query this file for its characteristics”. But I don't mind adding an intermediate class, let's call it `SerializationContext`, that gets populated by the `MachOFile` with its endianess and bit width properties and then that instance of `SerializationContext` is passed to the load command's `serialize` method. How does that sound?

lib/macho/load_commands.rb
UniqMartin: They will be complete in the sense that they will contain all relevant information. `cmdsize` is really only needed to read the load command from a byte stream (to know how much needs to be read). It doesn't carry any intrinsic value, as it can be computed from the payload it carries (at least for the commands we're primarily looking at right now, i.e. `LC_RPATH` and `LC_*_DYLIB`).

UniqMartin:
*Sorry for the long delay between my initial response and this concluding comment. I went away for dinner and then got a bit side-tracked with other stuff before returning here.*

I hope my comments didn't sound too harsh. I'm just feeling quite strongly about this and how this ought to be implemented. Here's a quick overview of how I imagine the life cycle of a created load command, maybe to make my code comments clearer:

1. The load command gets created out of nowhere. It has all the vital information (i.e. its ID and all the information specific to the particular load command class). It doesn't yet encode *any* information about its serialized representation, i.e. endianness, padding, and `cmdsize` (its padded size).

2. The load command gets serialized to a byte stream, probably by calling `MachOFile#add_command`, and then inserted into its target location in the Mach-O file. The load command being serialized here isn't altered at all. It's still in the state it had before it was serialized and can be serialized again into a different Mach-O file.

3. The load commands that were previously attached to the `MachOFile` instance are removed and completely re-read from the raw data that has been updated during the insertion of the new command. This results also in an instance that represents the newly inserted load command, including its `cmdsize` and an attached `view`.

  If re-reading all load commands gets too expensive, we could consider to only read (deserialize) the newly inserted command and to correct the view offsets of all load commands that follow it.

----

Some random notes:

- It would be good to make some of the refactoring that isn't immediately related to the feature at hand a separate commit (or ideally moved into a separate PR and merged first). I'm mainly talking about moving the constants into a separate file, which creates a big diff and makes it a bit harder than necessary to review the actual changes.

- I appreciate your effort to implement the serialization for as many commands as possible. I think it would be more efficient if you try to focus on the commands we're most interested in right now (`LC_RPATH` and `LC_*_DYLIB`) as these are sufficient to judge the overall design. Once we've settled on a design and made creation, serialization, etc. of those commands work (which might require more tweaks to the implementation) and are satisfied with how this works, it makes sense to implement serialization (and creation where it makes sense) for all other load command classes.

woodruffw:
Moving the discussion down here, since all but one of the remaining code comments revolve around one design decision.

> This will allow creation of a target-agnostic load command that can be then serialized to all Mach-O slices of a given fat binary, e.g. when adding an LC_RPATH command to a fat binary

How do you envision this API? I've been thinking of it like this:

```ruby
# creates N new rpath commands, one for each macho as macho.add_rpath("/foo")
fat_file.add_rpath("/foo")
```

While I think you're envisioning this (please correct me if I'm wrong!):

```ruby
# creates N serialized rpaths, one for each macho as macho.add_rpath(lc)
fat_file.add_rpath(lc)
```

Or perhaps (although I don't think either of us is thinking of this one):

```ruby
# creates N new rpath commands, one for each macho as macho.add_rpath(lc)
fat_file.add_rpath("/foo")
```

I'm convinced that your approach of making created load commands as agnostic is a good one, but I'm not sure that it's the best approach for the first API design. Conversely, I don't think my original approach is good at all for the second API design.

UniqMartin:
> How do you envision this API?

I think none of the options you presented fully captures what I had in mind (and it's somewhat comforting you aren't reading my mind yet 😀). My idea looks something like this:

- There's a method `add_rpath` that accepts a path string and lives in a separate module, let's call it `CreationHelpers`. This module is then mixed-in into both `FatFile` and `MachOFile` to provide both with this API. The `add_rpath` method sets up a target-agnostic instance of an `LC_RPATH` command with the given path and then invokes `add_command(lc)`.

- Depending on whether we're adding the command to a fat file or thin Mach-O file, this ends up calling `FatFile#add_command` or `MachOFile#add_command`. The latter does the straightforward thing, i.e. serializes the load command, inserts it into `raw_data`, and flushes/rebuilds any caches invalided by this. The former is even more straightforward, as it simply calls `add_command` for every of its Mach-O slices and just passes on the load command it has been given.

Does this make sense to you?

woodruffw:
> I hope my comments didn't sound too harsh. I'm just feeling quite strongly about this and how this ought to be implemented.

Not at all! I appreciate a critical review :smile: 

I think we've been imagining the lifecycle of a created load command completely differently, which is why we've come to different conclusions about what information they should be created with.

I've been imagining them as largely isolated from the user, created only within `MachOFile` instances. The exception to this would be a general `MachOFile#add_command` method, which would have to take a load command instance. Other than that, a method like `MachOFile#add_rpath` would behave identically to `MachOFile#dylib_id=` in accepting only a string as an argument.

I think this approach has advantages (like minimizing the number of arguments taken by `serialize` and creating more "complete" load commands), but yours does as well in the form of isolation and greater abstraction. You're also correct in that we definitely don't *need* fields like `cmdsize` to be populated in created load commands.

Ultimately, I'll defer to your judgement here. Both approaches will arrive at the same functionality with approximately the same work.

---

...and I just read your comment above (we seem to be on the same commenting schedule :smile:)

You've completely convinced me, using a mix-in is much clearer than what I thought you had in mind. It's also a lot conceptually cleaner than my original idea.

I'll go with your approach!

lib/macho/load_commands.rb
```diff
@@ -225,6 +282,30 @@ def to_i
...
+    class SerializationContext
+      # @return [Symbol] the endianness of the serialized string
+      attr_reader :endianness
+
+      # @return [Fixnum] the constant rounding value used to pad the serialized string
+      attr_reader :cmd_round
+
+      # @param macho [MachO::MachOFile] the file to contextualize
+      # @return [MachO::LoadCommand::SerializationContext] the resulting context
+      def self.context_for(macho)
```
woodruffw: Thoughts on this versus a `MachOFile#lc_context` method? I couldn't decide between the two, so I just flipped a coin :smile:

woodruffw:
I squashed out the *load_command_constants.rb* split that I made earlier, and applied your suggested structure. Please tell me what you think!

lib/macho/load_commands.rb
UniqMartin: That's fine by me. :+1:

lib/macho/load_commands.rb
```diff
@@ -225,6 +282,30 @@ def to_i
         @string_offset
       end
     end
+
+    # Represents the contextual information needed by a load command to
+    # serialize itself correctly into a binary string.
+    class SerializationContext
+      # @return [Symbol] the endianness of the serialized string
```
UniqMartin: I wonder whether the use of “string” here is going to cause confusion. I'd probably replace that with “load command”.

lib/macho/load_commands.rb
```diff
@@ -225,6 +282,30 @@ def to_i
...
+      attr_reader :endianness
+
+      # @return [Fixnum] the constant rounding value used to pad the serialized string
```
UniqMartin: I wonder whether the use of “string” here is going to cause confusion. I'd probably replace that with “load command”.

lib/macho/load_commands.rb
```diff
@@ -399,6 +480,19 @@ def initialize(view, cmd, cmdsize, name, timestamp, current_version, compatibili
...
+
+    # @param context [MachO::LoadCcommand::SerializationContext] the context
+    # @return [String] the serialized fields of the load command
+    # @api private
+    def serialize(context)
+      format = Utils.specialize_format(FORMAT, context.endianness)
+      lc_str = SIZEOF # a dylib command can only have one dylib
+      cmdsize = Utils.round(SIZEOF + name.to_s.bytesize + 1, context.cmd_round)
+      npad = cmdsize - SIZEOF - name.to_s.bytesize
+      padded_name = name.to_s + "\x00" * npad
```
UniqMartin: I'm basically fine with this code, but I'd like to see it being pulled out into a utility method instead of being replicated in every `serialize` method that involves an `lc_str`, as I suggested in a previous code comment: https://github.com/Homebrew/ruby-macho/pull/38#discussion_r70019349

lib/macho/load_commands.rb
```diff
@@ -181,6 +209,22 @@ def initialize(view, cmd, cmdsize)
...
+      CREATABLE_LOAD_COMMANDS.include?(LOAD_COMMANDS[cmd])
+    end
+
+    # @param context [MachO::LoadCommand::SerializationContext] the context
+    #  to serialize into
+    # @return [String, nil] the serialized fields of the load command, or nil
+    #  if the load command can't be serialized
+    # @api private
+    def serialize(context)
+      return unless serializable?
```
UniqMartin: I think in this case raising an exception is more appropriate than returning `nil`.

lib/macho/load_commands.rb
```diff
@@ -181,6 +209,22 @@ def initialize(view, cmd, cmdsize)
...
+      format = Utils.specialize_format(FORMAT, context.endianness)
+      [cmd, cmdsize].pack(format)
```
UniqMartin: Is this intended as a generic implementation that works for load commands that only contain a fixed set of fields and have a fixed size? If so, `cmdsize` still needs to be populated with `SIZEOF` (or simply replaced with it in this line).

lib/macho/load_commands.rb
```diff
@@ -171,6 +181,24 @@ def self.new_from_bin(view)
       self.new(view, *bin.unpack(format))
     end
 
+    # Creates a new (viewless) command corresponding to the symbol provided
+    # @param cmd_sym [Symbol] the symbol of the load command being created
+    # @param args [Array] the arguments for the load command being created
+    # @note this does not currently work. DO NOT USE IT.
```
UniqMartin: This note might be no longer true.

lib/macho/load_commands.rb
```diff
@@ -205,14 +249,27 @@ def to_s
...
+
+        if view
+          lc_str_abs = view.offset + lc_str
+          lc_end = view.offset + lc.cmdsize - 1
+          @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
+          @string_offset = lc_str
+        else
+          # setting string_offset to the size of the command assumes that there
+          # is only one lc_str in the command. this is presently true for all
+          # load commands, but is not guaranteed in the future.
```
UniqMartin: I think this comment (no longer) belongs here. It's also not fully accurate either, as there's `prebound_dylib_command` that hosts two instances of an `lc_str`, thus multiple `lc_str`s are not just a future thing.

test/test_serialize_load_commands.rb
```diff
@@ -0,0 +1,245 @@
...
+  def test_serialize_load_dylib
+    filenames = SINGLE_ARCHES.map { |a| fixture(a, "hello.bin") }
+
+    filenames.each do |filename|
+      file = MachO::MachOFile.new(filename)
+      ctx = MachO::LoadCommand::SerializationContext.context_for(file)
+      lc = file[:LC_LOAD_DYLIB].first
+      lc2 = MachO::LoadCommand.create(:LC_LOAD_DYLIB, lc.name.to_s,
+        lc.timestamp, lc.current_version, lc.compatibility_version)
+      blob = lc.view.raw_data[lc.view.offset, lc.cmdsize]
```
UniqMartin: See also my earlier code comment that might have went unnoticed and is now collapsed by default: https://github.com/Homebrew/ruby-macho/pull/38#discussion_r70019985

UniqMartin:
> I squashed out the *load_command_constants.rb* split that I made earlier,

Just to be clear, I'm not opposed to relocating those constant into their own file. Just not as part of this PR. But that's a simple enough change to make afterwards without much hassle.

> and applied your suggested structure. Please tell me what you think!

There are still some quirks and implementation details that need to be addressed (I made a few code comments to that end), but overall that's basically what I was hoping to arrive at.

lib/macho/load_commands.rb
woodruffw: Sorry that I missed that command. A utility method sounds good to me!

UniqMartin: Sorry, couldn't resist 🙈: It's a comment, I'm not *command*ing you just yet. :joy:

woodruffw:
> Just to be clear, I'm not opposed to relocating those constant into their own file. Just not as part of this PR.

Yep, I understood. I'll include that change in a general refactor in a few days (moving constants into their own files).

lib/macho/load_commands.rb
woodruffw: > Is this intended as a generic implementation that works for load commands that only contain a fixed set of fields and have a fixed size?

Yep. I'll fix this.

lib/macho/load_commands.rb
woodruffw: Agreed, how about `LoadCommandNotSerializableError`?

lib/macho/load_commands.rb
woodruffw: Whoops, that was actually a typo :smile:

lib/macho/load_commands.rb
UniqMartin: Sounds good!

test/test_serialize_load_commands.rb
woodruffw: I agree, but having it be called `raw_data` might be confusing in light of `view.raw_data`. It's also a little confusing in terms of the presence of `serialize`, although not as much so.

Perhaps `raw_lc` (or just `raw`)?

lib/macho/load_commands.rb
woodruffw: Applied!

test/test_serialize_load_commands.rb
UniqMartin: Yeah, you're right. Calling it `raw_data` could be confusing. Right now I'm thinking maybe `raw_from_view` to make it very clear that this has nothing to do with `serialize`, but right now it's probably best to ignore this issue and leave it as-is. (The repeated occurrence of this piece of code isn't bothering me enough to warrant stalling this PR.)

lib/macho/utils.rb
```diff
@@ -11,6 +11,13 @@ def self.round(value, round)
       value
     end
 
+    # @param size [Fixnum] the pre-padded size
```
UniqMartin: I think “unpadded size” should be clearer; “pre-padded” sounds to me like there's already some padding in place.

lib/macho/macho_file.rb
```diff
@@ -64,6 +64,10 @@ def magic64?
       Utils.magic64?(header.magic)
     end
 
+    def cmd_round
```
UniqMartin: It's a bit of a bike shedding topic I guess, but I'm not entirely happy with the terminology. While what is happening at the end is that the size of a command is rounded to a multiple of a certain number by padding the end with some null bytes, the number being returned here is probably better named `alignment`. That's basically the whole point of padding the load commands, namely to ensure that every load command starts at a memory address that is a multiple of either 4 or 8, because for certain architectures this alignment is important to be able to actually access the individual fields of those load command structures. (I think this is also more recognizable terminology for people familiar with C-like languages.)

Sorry for the long introduction, I just tried to provide enough context. **TL;DR** It would be nice to replace all newly introduced instances of `cmd_round` or `round` with `alignment`.

UniqMartin:
Two final comments, but nothing that involves any fundamental changes. Feel free to merge without further feedback from me once addressed. If necessary, we can fix stuff we might have overlooked here as it comes up, but I think the design is sound as-is.

Let's get this shipped so we can start reaping the benefits! Thanks for bearing with me! :bow:

lib/macho/macho_file.rb
woodruffw: Sounds good! I've just been going with `cmd_round` and `round` because of the `round` function in libmacho. `alignment` is much more descriptive.

UniqMartin: It reads a lot nicer now (at least to me)! Thanks!

UniqMartin:
I was happy with the rest, so I just went ahead and merged this. :tada: Thanks for following through with the extensive modifications I requested since your first draft!

PR #39, opened 2016-07-08
Add rpath changing to tools, tests.

woodruffw:
`Tools.add_rpath` and `Tools.delete_rpath` remain unimplemented, for the time being. I'll be implementing those in another PR.

The first shouldn't be too hard once creation/serialization of LCs is completed, and the second is just a matter of finding the right LC and removing it entirely (and updating `sizeofcmds`).


UniqMartin:
:+1:

PR #42, opened 2016-07-13
macho_file: Refactor MachOFile#sections.

woodruffw:
Takes advantage of a case to determine the correct class.

cc @UniqMartin

PR #43, opened 2016-07-13
macho_file: General refactoring.

woodruffw:
Replaces #42.

cc @UniqMartin

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -188,9 +188,7 @@ module MachO
     #  file.dylib_id # => 'libBar.dylib'
     # @return [String, nil] the Mach-O's dylib ID
     def dylib_id
-      if !dylib?
-        return nil
-      end
+      return unless dylib?
 
       dylib_id_cmd = command(:LC_ID_DYLIB).first
 
@@ -204,13 +202,8 @@ module MachO
     # @return [void]
     # @raise [ArgumentError] if `new_id` is not a String
     def dylib_id=(new_id)
-      if !new_id.is_a?(String)
-        raise ArgumentError.new("argument must be a String")
-      end
-
-      if !dylib?
-        return nil
-      end
+      raise ArgumentError.new("new ID must be a String") unless new_id.is_a?(String)
+      return unless dylib?
 
       dylib_cmd = command(:LC_ID_DYLIB).first
       old_id = dylib_id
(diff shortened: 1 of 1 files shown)


lib/macho/macho_file.rb
```diff
@@ -204,13 +202,8 @@ def dylib_id
...
     # @raise [ArgumentError] if `new_id` is not a String
     def dylib_id=(new_id)
-      if !new_id.is_a?(String)
-        raise ArgumentError.new("argument must be a String")
-      end
-
-      if !dylib?
-        return nil
-      end
+      raise ArgumentError.new("argument must be a String") unless new_id.is_a?(String)
```
UniqMartin: That's a style issue we can discuss and that I haven't brought up so far, but I'd prefer to split this across several lines to make the overall line length shorter (to stay within a with of 80 columns), i.e.:

```ruby
unless new_id.is_a?(String)
  raise ArgumentError.new("argument must be a String")
end
```

(And now that you're touching this code, “argument” feels a tad too generic in the error message.)

lib/macho/macho_file.rb
```diff
@@ -269,26 +262,23 @@ def change_rpath(old_path, new_path)
...
 
       segment.nsects.times do
-        if segment.is_a? SegmentCommand
-          sections << Section.new_from_bin(endianness, @raw_data.slice(offset, Section.bytesize))
-          offset += Section.bytesize
-        else
-          sections << Section64.new_from_bin(endianness, @raw_data.slice(offset, Section64.bytesize))
-          offset += Section64.bytesize
-        end
+        sections << section_klass.new_from_bin(endianness, @raw_data.slice(offset, section_klass.bytesize))
```
UniqMartin: Split this into two lines for shorter lines and better readability:

```ruby
section_bin = @raw_data[offset, section_klass.bytesize]
sections << section_klass.new_from_bin(endianness, section_bin)
```

UniqMartin:
Two style suggestions, but otherwise :+1:.

lib/macho/macho_file.rb
woodruffw: My preference has generally been to turn expressions like this into single lines if they're only *slightly* longer than 80 columns, the rationale being that it keeps the method's exclusion logic as short as possible.

That being said, I don't feel strongly about it as a style - I'm happy to revert it if you prefer the block form (for >  80 lines).



> (And now that you're touching this code, “argument” feels a tad too generic in the error message.)

Agreed, changing it to "new ID" :smile:

UniqMartin: I guess it's more consistent with most of the other code in the project, thus I'm fine with leaving it as-is and revisiting at a later stage.

UniqMartin:
Thanks! Went ahead and merged this as it looked ready to me. 😄

PR #44, opened 2016-07-14
Rpath addition.

woodruffw:
Broken off from #40. ~~WIP until #38 is merged.~~

cc @UniqMartin

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -170,6 +170,30 @@ module MachO
 
     alias :[] :command
 
+    # Add a new load command to the Mach-O.
+    # @param lc [MachO::LoadCommand] the load command being added
+    # @return [void]
+    # @raise [MachO::HeaderPadError] if the new command exceeds the header pad buffer
+    # @note This is public, but methods like {#add_rpath} should be preferred.
+    def add_command(lc)
+      context = LoadCommand::SerializationContext.context_for(self)
+      cmd_raw = lc.serialize(context)
+      new_sizeofcmds = sizeofcmds + cmd_raw.bytesize
+
+      if header.class.bytesize + new_sizeofcmds > low_fileoff
+        raise HeaderPadError.new(@filename)
+      end
+
+      # update the raw data to reflect the new command
+      set_ncmds(ncmds + 1)
+      set_sizeofcmds(new_sizeofcmds)
+      @raw_data[header.class.bytesize + sizeofcmds, cmd_raw.bytesize] = cmd_raw
+
+      # synchronize fields with the raw data
+      @header = get_mach_header
+      @load_commands = get_load_commands
+    end
+
     # All load commands responsible for loading dylibs.
     # @return [Array<MachO::DylibCommand>] an array of DylibCommands
     def dylib_load_commands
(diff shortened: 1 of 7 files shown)


woodruffw:
I went ahead and implemented `MachOFile#add_command` in this PR, since `add_rpath`'s behavior is a subset of it.

test/test_tools.rb
```diff
@@ -146,11 +146,52 @@ def test_change_rpath_fat
...
+      MachO::Tools.add_rpath(actual, "/foo/bar/baz")
+
+      original = MachO::MachOFile.new(filename)
+      modified = MachO::MachOFile.new(actual)
+
+      assert_operator modified.ncmds, :>, original.ncmds
+      assert_operator modified.sizeofcmds, :>, original.sizeofcmds
+      assert_operator modified.rpaths.size, :>, original.rpaths.size
+      refute_includes original.rpaths, "/foo/bar/baz"
+      assert_includes modified.rpaths, "made_up_path"
```
woodruffw: Whoops, this should also be `"/foo/bar/baz"`. I'll fix this when I'm home.

test/test_tools.rb
```diff
@@ -146,11 +146,52 @@ def test_change_rpath_fat
...
+
+    groups.each do |filename, actual|
+      FileUtils.cp filename, actual
+      MachO::Tools.add_rpath(actual, "/foo/bar/baz")
+
+      original = MachO::FatFile.new(filename)
+      modified = MachO::FatFile.new(actual)
+
+      refute_includes original.rpaths, "/foo/bar/baz"
+      assert_includes modified.rpaths, "made_up_path"
```
woodruffw: Same here.

test/test_tools.rb
UniqMartin: So what's wrong with the test if this mistake didn't cause the test run to fail?

woodruffw: It's just testing the wrong string - `"made_up_path"` was already in each file, `"/foo/bar/baz"` is the one that was added and should be tested for.

UniqMartin: Oh, right! This is all about addition so of course the pre-existing RPATH is still included.

woodruffw: Yep, just a careless copy-paste :smile:

lib/macho/macho_file.rb
```diff
@@ -170,6 +170,33 @@ def command(name)
...
     alias :[] :command
 
+    # Add a new load command to the Mach-O.
+    # @param lc [MachO::LoadCommand] the load command being added
+    # @return [void]
+    # @raise [MachO::LoadCommandNotSerializableError] if the command can't be serialized
+    # @raise [MachO::HeaderPadError] if the new command exceeds the header pad buffer
+    # @note This is public, but methods like {#add_rpath} should be preferred.
+    def add_command(lc)
+      raise LoadCommandNotSerializableError.new(LOAD_COMMANDS[lc]) unless lc.serializable?
```
UniqMartin: Is this check actually needed? I think it is redundant as the `LoadCommand#serialize` invocation below will perform the exact same check.

UniqMartin:
One minor remark, but otherwise this looks good to me. :+1:

lib/macho/macho_file.rb
woodruffw: Nope, just an oversight. I'll fix it.

woodruffw:
> One minor remark, but otherwise this looks good to me. :+1:

Great! Merging now...

UniqMartin:
Awesome! :tada:

I think the next steps would/should be:

- Re-implement `MachOFile#change_rpath` to use load command serialization.
- Re-implement `MachOFile#change_dylib_id` to use load command serialization.
- Re-implement `MachOFile#change_install_name` to use load command serialization.
- Drop no longer needed `MachOFile#set_lc_str_in_cmd`.

Am I right? Anything else that needs to be addressed beforehand?

woodruffw:
Yep, that sounds right. I'll open a PR for these tonight :smile:

Issue #41, opened 2016-07-09
Duplicated LC modification logic


woodruffw:
Currently, `MachOFile#change_rpath` allows a duplicate rpath to be created, since it does not check the names of rpaths already present in the file.

This behavior isn't *incorrect* in the sense that it's also what `install_name_tool` does, but it's unintuitive and will likely confuse users who expect a given rpath to occur only once. More to the point, following `install_name_tool`s behavior will allow a user to create a binary that triggers another unintuitive behavior in `install_name_tool` (namely that rpath deletion only deletes the *first* matching rpath, not *all* matching rpaths).

ref https://github.com/Homebrew/ruby-macho/pull/40#discussion_r70173115
cc @UniqMartin

UniqMartin:
I think this is a good little issue to tackle once you have implemented RPATH addition and are waiting for feedback/review from me (as I sadly don't always manage to catch up with your work on a daily basis). It would certainly be nice to have all the RPATH-related features cleaned up and operating somewhat consistently.

woodruffw:
> It would certainly be nice to have all the RPATH-related features cleaned up and operating somewhat consistently.

UniqMartin:
I guess now would be a good time to address this little issue when modifying an existing RPATH and maybe spend a few moments to think whether any of this could be relevant in the context of the `#dylib_id=` or `#change_install_name` methods.

woodruffw:
Yep!

Unlike with rpaths, dylibs/dylib ids have associated version information that might  be relevant if we decide to delete duplicated dylib commands. I'll have to do some tests on the my system libraries/homebrew-installed libs, but based on the fact that `install_name_tool` doesn't ask for the dylib's compatibility version we can probably assume that duplicated dylibs are treated similarly to duplicated rpaths.

woodruffw:
I found a [CMake thread](https://cmake.org/pipermail/cmake/2014-October/058859.html) from 2014 that suggests that `install_name_tool` used to corrupt Mach-Os containing duplicate rpaths. It seems like that behavior was quietly fixed at some point in the last two years.

This "succeeds" for me (in the sense that it doesn't corrupt the binary, but leaves one duplicate rpath behind) on 10.11.5. Do you happen to have any older OS X versions to compare with?

```bash
echo "void foo() {}" > lib.c
clang -dynamiclib -o libtest.dylib -Wl,-rpath,/usr/lib -Wl,-rpath,/usr/lib lib.c
[...]

PR #45, opened 2016-07-16
Remove MachOFile#set_lc_str_in_cmd in favor of creation/serialization

woodruffw:
Additionally adds `MachOFile#populate_fields`, `MachOFile#replace_command`, `MachOFile#insert_command`, and `MachOFile#delete_command`, and generalizes the case where we delete multiple commands in batch and defer field synchronization until the end.

In the spirit of #41, do we want to prevent `MachOFile#change_install_name` from producing duplicate dylibs? My suspicion is that we do, again against the behavior of `install_name_tool`.

cc @UniqMartin

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -36,16 +36,14 @@ module MachO
 
       @filename = filename
       @raw_data = File.open(@filename, "rb") { |f| f.read }
-      @header = get_mach_header
-      @load_commands = get_load_commands
+      populate_fields
     end
 
     # @api private
     def initialize_from_bin(bin)
       @filename = nil
       @raw_data = bin
-      @header = get_mach_header
-      @load_commands = get_load_commands
+      populate_fields
     end
 
     # The file's raw Mach-O data.
@@ -170,26 +168,104 @@ module MachO
 
     alias :[] :command
 
-    # Add a new load command to the Mach-O.
-    # @param lc [MachO::LoadCommand] the load command being added
-    # @return [void]
+    # Inserts a load command at the given offset.
+    # @param offset [Fixnum] the offset to insert at
+    # @param lc [MachO::LoadCommand] the load command to insert
+    # @param options [Hash]
+    # @option options [Boolean] :repopulate (true) whether or not to repopulate
+    #  the instance fields
+    # @raise [MachO::OffsetInsertionError] if the offset is not in the load command region
     # @raise [MachO::HeaderPadError] if the new command exceeds the header pad buffer
-    # @note This is public, but methods like {#add_rpath} should be preferred.
-    def add_command(lc)
+    # @note Calling this method with an arbitrary offset in the load command
+    #  region **will leave the object in an inconsistent state**.
+    def insert_command(offset, lc, options = {})
       context = LoadCommand::SerializationContext.context_for(self)
       cmd_raw = lc.serialize(context)
+
+      if offset < header.class.bytesize || offset + cmd_raw.bytesize > low_fileoff
+        raise OffsetInsertionError.new(offset)
+      end
+
       new_sizeofcmds = sizeofcmds + cmd_raw.bytesize
 
       if header.class.bytesize + new_sizeofcmds > low_fileoff
         raise HeaderPadError.new(@filename)
       end
 
-      # update the raw data to reflect the new command
+      # update Mach-O header fields to account for inserted load command
       set_ncmds(ncmds + 1)
       set_sizeofcmds(new_sizeofcmds)
-      @raw_data[header.class.bytesize + sizeofcmds, cmd_raw.bytesize] = cmd_raw
 
-      # synchronize fields with the raw data
+      @raw_data.insert(offset, cmd_raw)
+      @raw_data.slice!(header.class.bytesize + new_sizeofcmds, cmd_raw.bytesize)
+
+      populate_fields if options.fetch(:repopulate, true)
+    end
+
+    # Replace a load command with another command in the Mach-O, preserving location.
+    # @param old_lc [MachO::LoadCommand] the load command being replaced
+    # @param new_lc [MachO::LoadCommand] the load command being added
+    # @return [void]
+    # @raise [MachO::HeaderPadError] if the new command exceeds the header pad buffer
+    # @see {#insert_command}
+    # @note This is public, but methods like {#dylib_id=} should be preferred.
+    def replace_command(old_lc, new_lc)
+      context = LoadCommand::SerializationContext.context_for(self)
+      cmd_raw = new_lc.serialize(context)
+      new_sizeofcmds = sizeofcmds + cmd_raw.bytesize - old_lc.cmdsize
+      if header.class.bytesize + new_sizeofcmds > low_fileoff
+        raise HeaderPadError.new(@filename)
+      end
+
+      delete_command(old_lc)
+      insert_command(old_lc.view.offset, new_lc)
+    end
+
+    # Appends a new load command to the Mach-O.
+    # @param lc [MachO::LoadCommand] the load command being added
+    # @param options [Hash]
+    # @option options [Boolean] :repopulate (true) whether or not to repopulate
+    #  the instance fields
+    # @return [void]
+    # @see {#insert_command}
+    # @note This is public, but methods like {#add_rpath} should be preferred.
+    #  Setting `repopulate` to false **will leave the instance in an
+    #  inconsistent state** unless {#populate_fields} is called **immediately**
+    #  afterwards.
+    def add_command(lc, options = {})
+      insert_command(header.class.bytesize + sizeofcmds, lc, options)
+    end
+
+    # Delete a load command from the Mach-O.
+    # @param lc [MachO::LoadCommand] the load command being deleted
+    # @param options [Hash]
+    # @option options [Boolean] :repopulate (true) whether or not to repopulate
+    #  the instance fields
+    # @return [void]
+    # @note This is public, but methods like {#delete_rpath} should be preferred.
+    #  Setting `repopulate` to false **will leave the instance in an
+    #  inconsistent state** unless {#populate_fields} is called **immediately**
+    #  afterwards.
+    def delete_command(lc, options = {})
+      @raw_data.slice!(lc.view.offset, lc.cmdsize)
+
+      # update Mach-O header fields to account for deleted load command
+      set_ncmds(ncmds - 1)
+      set_sizeofcmds(sizeofcmds - lc.cmdsize)
+
+      # pad the space after the load commands to preserve offsets
+      null_pad = "\x00" * lc.cmdsize
+      @raw_data.insert(header.class.bytesize + sizeofcmds - lc.cmdsize, null_pad)
+
+      populate_fields if options.fetch(:repopulate, true)
+    end
+
+    # Populate the instance's fields with the raw Mach-O data.
+    # @return [void]
+    # @note This method is public, but should (almost) never need to be called.
+    #  The exception to this rule is when methods like {#add_command} and
+    #  {#delete_command} have been called with `repopulate = false`.
+    def populate_fields
       @header = get_mach_header
       @load_commands = get_load_commands
     end
@@ -233,10 +309,13 @@ module MachO
       raise ArgumentError.new("new ID must be a String") unless new_id.is_a?(String)
       return unless dylib?
 
-      dylib_cmd = command(:LC_ID_DYLIB).first
-      old_id = dylib_id
+      old_lc = command(:LC_ID_DYLIB).first
+      raise DylibIdMissingError.new unless old_lc
 
-      set_name_in_dylib(dylib_cmd, old_id, new_id)
+      new_lc = LoadCommand.create(:LC_ID_DYLIB, new_id, old_lc.timestamp,
+        old_lc.current_version, old_lc.compatibility_version)
+
+      replace_command(old_lc, new_lc)
     end
 
     # All shared libraries linked to the Mach-O.
@@ -257,10 +336,13 @@ module MachO
     # @return [void]
     # @raise [MachO::DylibUnknownError] if no shared library has the old name
     def change_install_name(old_name, new_name)
-      dylib_cmd = dylib_load_commands.find { |d| d.name.to_s == old_name }
-      raise DylibUnknownError.new(old_name) if dylib_cmd.nil?
+      old_lc = dylib_load_commands.find { |d| d.name.to_s == old_name }
+      raise DylibUnknownError.new(old_name) if old_lc.nil?
 
-      set_name_in_dylib(dylib_cmd, old_name, new_name)
+      new_lc = LoadCommand.create(:LC_LOAD_DYLIB, new_name,
+        old_lc.timestamp, old_lc.current_version, old_lc.compatibility_version)
+
+      replace_command(old_lc, new_lc)
     end
 
     alias :change_dylib :change_install_name
@@ -279,10 +361,12 @@ module MachO
     # @return [void]
     # @raise [MachO::RpathUnknownError] if no such old runtime path exists
     def change_rpath(old_path, new_path)
-      rpath_cmd = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
-      raise RpathUnknownError.new(old_path) if rpath_cmd.nil?
+      old_lc = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
+      raise RpathUnknownError.new(old_path) if old_lc.nil?
 
-      set_path_in_rpath(rpath_cmd, old_path, new_path)
+      new_lc = LoadCommand.create(:LC_RPATH, new_path)
+
+      replace_command(old_lc, new_lc)
     end
 
     # Add the given runtime path to the Mach-O.
@@ -316,29 +400,11 @@ module MachO
       # sequential order, but let's not rely on this behavior
       rpath_cmds.sort_by! { |lc| lc.view.offset }
 
-      # calculate the size of all commands pending deletion
-      deleted_size = rpath_cmds.map(&:cmdsize).inject(0, :+)
+      # delete the commands in reverse order, offset descending. this
+      # allows us to defer (expensive) field population until the very end
+      rpath_cmds.reverse_each { |cmd| delete_command(cmd, repopulate: false) }
 
-      # this behavior differs from that of install_name_tool, which
-      # will only delete the first matching rpath, instead of all matching ones
-      rpath_cmds.reverse_each do |rpath_cmd|
-        # slice the rpath out of the raw data
-        @raw_data.slice!(rpath_cmd.view.offset, rpath_cmd.cmdsize)
-      end
-
-      # decrease the number of load commands by the number deleted
-      set_ncmds(ncmds - rpath_cmds.size)
-
-      # decrease the total size of commands by the size of the deleted commands
-      set_sizeofcmds(sizeofcmds - deleted_size)
-
-      # pad the space after the load commands to preserve offsets
-      null_pad = "\x00" * deleted_size
-      @raw_data.insert(header.class.bytesize + sizeofcmds, null_pad)
-
-      # synchronize fields with the raw data
-      @header = get_mach_header
-      @load_commands = get_load_commands
+      populate_fields
     end
 
     # All sections of the segment `segment`.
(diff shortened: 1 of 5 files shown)


lib/macho/macho_file.rb
```diff
@@ -189,7 +189,38 @@ def add_command(lc)
...
+      @raw_data.insert(header.class.bytesize + sizeofcmds, null_pad)
+
+      synchronize_fields! if synchronize
+    end
+
+    # Synchronize the instance's fields with the raw Mach-O data.
+    # @return [void]
+    # @note This method is public, but should (almost) never need to be called.
+    #  The exception to this rule is when methods like {#add_command} and
+    #  #{delete_command} have been called with `synchronization_false`.
```
UniqMartin: I suspect this should be `{#delete_command}` (misplaced octothorpe).

lib/macho/macho_file.rb
```diff
@@ -189,7 +189,38 @@ def add_command(lc)
...
+    def synchronize_fields!
```
UniqMartin: I think we can drop the bang (`!`) at the end and I'm afraid this method name could be confusing. When I'm thinking of synchronization, I'm thinking more about flushing the changes that are only in memory to disk (or in this case to raw data), i.e. exactly in the opposite direction.

How do you feel about naming this `reload_fields`, `reload_from_raw_data` (a bit uglier but also clearer), or something similar (better ideas for a name are more than welcome)?

lib/macho/macho_file.rb
woodruffw: Whoops, fixing this...

lib/macho/macho_file.rb
```diff
@@ -233,10 +264,12 @@ def dylib_id=(new_id)
       raise ArgumentError.new("new ID must be a String") unless new_id.is_a?(String)
       return unless dylib?
 
-      dylib_cmd = command(:LC_ID_DYLIB).first
-      old_id = dylib_id
+      old_cmd = command(:LC_ID_DYLIB).first
+      new_cmd = LoadCommand.create(:LC_ID_DYLIB, new_id, old_cmd.timestamp,
+        old_cmd.current_version, old_cmd.compatibility_version)
```
UniqMartin: What if there is no `LC_ID_DYLIB` load command in the file despite it being a dylib? (I think I once came across a file like that, though normally a dylib of course would always have an ID.)

lib/macho/macho_file.rb
woodruffw: > When I'm thinking of synchronization, I'm thinking more about flushing the changes that are only in memory to disk (or in this case to raw data), i.e. exactly in the opposite direction.

That makes sense, especially considering that the only other bang-method in `MachOFile` is `write!`. I'll remove the `!`.

I went with `synchronize_fields` because I ended up using it in the constructor(s) as well (where I think it would sound a little strange to call `reload_fields` or something similar). That being said, `reload_fields` or perhaps `refresh_fields` would be fine by me if we revert to populating `@header` and `@load_commands` explicitly in the constructors.

lib/macho/macho_file.rb
```diff
@@ -233,10 +264,12 @@ def dylib_id=(new_id)
...
 
-      set_name_in_dylib(dylib_cmd, old_id, new_id)
+      add_command(new_cmd)
+      delete_command(old_cmd)
```
UniqMartin: I applaud the simplicity created by this, but I think we might want to additionally implement a `replace_command` method to avoid reshuffling commands just because they are being modified.

This shouldn't be an issue here, but it can definitely be problematic for the family of `LC_LOAD_DYLIB` commands where load order can also influence against which dylib a symbol is resolved and also for `LC_RPATH` commands as this can affect what library ends up being loaded.

Another problem in this specific case is that this can fail even if there is enough space for the changed dylib ID, simply because the new one is added before the old one is removed, thus temporarily there needs to be space for both load commands, which is more than would be strictly necessary.

lib/macho/macho_file.rb
woodruffw: My intuition is that we should probably create one if it's missing, although I have an equally strong intuition that any dylib missing its ID is a huge warning sign and should probably be interpreted as an incorrectly formatted Mach-O.

I'll remove the ID from one of our test dylibs and see how OS X behaves with it and defer to that behavior :smile:

lib/macho/macho_file.rb
UniqMartin: The constructor should definitely also use this method and avoid duplicating code. How about naming this `populate_fields` or maybe even `parse_raw_data`? This should work (on a language level) both for the first invocation as well as for subsequent invocations due to modifications being made to the load commands.

lib/macho/macho_file.rb
woodruffw: > This shouldn't be an issue here, but it can definitely be problematic for the family of LC_LOAD_DYLIB commands where load order can also influence against which dylib a symbol is resolved and also for LC_RPATH commands as this can affect what library ends up being loaded.

Ah, this was a huge oversight on my part :sweat_smile:

lib/macho/macho_file.rb
woodruffw: I think `populate_fields` sounds good :+1:

lib/macho/macho_file.rb
UniqMartin: Creating one if it's missing is probably not a good idea, particularly since we wouldn't know to what values we should set `timestamp`, `current_version`, and `compatibility_version`.

But raising an exception if there's none sounds like a reasonable thing to do and would avoid a hard to interpret failure when trying to access the above mentioned fields of `nil`.

lib/macho/macho_file.rb
UniqMartin: Another idea that just popped into my head: `decode_raw_data`.

lib/macho/macho_file.rb
woodruffw: > But raising an exception if there's none sounds like a reasonable thing to do and would avoid a hard to interpret failure when trying to access the above mentioned fields of nil.

Agreed - How does `DylibIdMissingError` sound?

lib/macho/macho_file.rb
woodruffw: My only real issue with `parse_raw_data`/`decode_raw_data` is that they sound like they return parsed/decoded raw data, rather than update the internal state of the object.

lib/macho/macho_file.rb
UniqMartin: I think you can retain most of the current logic if you make sure to always first delete the old and then insert the new command (thus keeping it a two-step process). This would require an `insert_command` method that could allow injecting a load command at a given offset. Then `add_command` could become a special case of `insert_command` where the insertion point just happens to be right after the last load command.

woodruffw: Since `replace_command` will take two `LoadCommand`s as arguments, we should probably prevent the user from replacing one class of command with another, right? Thoughts on `LoadCommandMismatchError`?

lib/macho/macho_file.rb
UniqMartin: > Agreed - How does `DylibIdMissingError` sound?

Prefect! 👌

lib/macho/macho_file.rb
UniqMartin: Yep, that's a valid concern! If you're happy with `populate_fields`, then I'm happy, too. I guess this means the optional argument to `add_command`, `delete_command` would be named something like `repopulate`?

woodruffw: Yup, I had it as `populate` but `repopulate` is much more accurate :smile:

lib/macho/macho_file.rb
UniqMartin: I'm not sure we need to be this strict. But see my other comment above on why we might not need a `replace_command` method and instead can retain the two-step procedure if we used insertion instead of addition.

lib/macho/macho_file.rb
```diff
@@ -189,7 +189,38 @@ def add_command(lc)
...
+    #  Setting `synchronize` to false **will break the instance** unless
+    #  {#synchronize_fields} is called **immediately** afterwards.
+    def delete_command(lc, synchronize = true)
+      @raw_data.slice!(lc.view.offset, lc.cmdsize)
+
+      # decrease the number of load commands by the number deleted
+      set_ncmds(ncmds - 1)
+
+      # decrease the total size of commands by the size of the deleted commands
+      set_sizeofcmds(sizeofcmds - lc.cmdsize)
```
UniqMartin: I think you can be a bit less verbose with this and the previous non-empty/non-comment line:

```ruby
# Update Mach-O header fields to account for deleted load command.
set_ncmds(ncmds - 1)
set_sizeofcmds(sizeofcmds - lc.cmdsize)
```

(The rest is already evident from the code itself and the surrounding method.)

lib/macho/macho_file.rb
```diff
@@ -189,7 +189,38 @@ def add_command(lc)
...
+
+      # pad the space after the load commands to preserve offsets
+      null_pad = "\x00" * lc.cmdsize
+      @raw_data.insert(header.class.bytesize + sizeofcmds, null_pad)
```
UniqMartin: This isn't technically broken, but it doesn't align with the comment above this block, as this is inserting zero bytes at the previous end of the load command list, not at the point where the end of the list is now that the specified command has been deleted.

lib/macho/macho_file.rb
```diff
@@ -189,7 +189,38 @@ def add_command(lc)
...
+
+      synchronize_fields if synchronize
+    end
+
+    # Synchronize the instance's fields with the raw Mach-O data.
+    # @return [void]
+    # @note This method is public, but should (almost) never need to be called.
+    #  The exception to this rule is when methods like {#add_command} and
+    #  {#delete_command} have been called with `synchronization_false`.
```
UniqMartin: “[…] called with `synchronization_false`”?

lib/macho/macho_file.rb
```diff
@@ -189,7 +189,38 @@ def add_command(lc)
...
-      # synchronize fields with the raw data
+      synchronize_fields if synchronize
+    end
+
+    # Delete a load command from the Mach-O.
+    # @param lc [MachO::LoadCommand] the load command being deleted
+    # @param synchronize [Boolean] whether or not to synchronize the instance fields
+    # @return [void]
+    # @note This is public, but methods like {#delete_rpath} should be preferred.
+    #  Setting `synchronize` to false **will break the instance** unless
```
UniqMartin: Let's rephrase “will break the instance” to “will leave the instance in an inconsistent state”. I think this more accurately describes what will be happening.

test/test_fat.rb
```diff
@@ -277,7 +277,10 @@ def test_change_dylib_id
 
       file.write(actual)
 
-      assert equal_sha1_hashes(actual, expected)
+      act = MachO::FatFile.new(actual)
+      exp = MachO::FatFile.new(expected)
+
+      assert_equal exp.dylib_id, act.dylib_id
```
UniqMartin: Just a reminder: You should be able to revert this and similar changes in the other test methods once `insert_command` is implemented and load command order is preserved again.

lib/macho/macho_file.rb
woodruffw: Maybe I'm understanding this incorrectly, but I don't think `add_command`/command modification generalize well into `insert_command`. In the `add_command` case we do a straight replacement of the null bytes at the end of the load commands, while in the modification case we do an insertion of serialized data at an offset and then pad/depad at the end of the load commands. In order to reconcile these two, we'd have to re-calculate our position in the file and thereby spoil the abstraction of the offset.

(There's a good chance I'm understanding this incorrectly - I'm pretty tired)

lib/macho/macho_file.rb
woodruffw: Yep, those were just leftovers from the older implementation. I'll update this.

UniqMartin:
> In the spirit of #41, do we want to prevent `MachOFile#change_install_name` from producing duplicate dylibs? My suspicion is that we do, again against the behavior of `install_name_tool`.

Yes, definitely. I'd personally prefer if this PR was purely about getting rid of `set_lc_str_in_cmd` in favor of command deletion and insertion and the semantic change in handling duplicates (and possibly other changes to the behavior) was postponed to a follow-up PR.

lib/macho/macho_file.rb
woodruffw: Whoops, good catch :+1:

lib/macho/macho_file.rb
woodruffw: Typo, I've corrected it to `repopulate = false` locally :smile:

test/test_fat.rb
woodruffw: Yep, I remember :wink:

woodruffw:
> Yes, definitely. I'd personally prefer if this PR was purely about getting rid of set_lc_str_in_cmd in favor of command deletion and insertion and the semantic change in handling duplicates (and possibly other changes to the behavior) was postponed to a follow-up PR.

Sounds good to me!

lib/macho/macho_file.rb
UniqMartin: Yes, currently we do a straight replacement of the zero bytes in `add_command`. But while it is a bit less efficient, that's effectively the same as inserting the serialized new command at the end of the current command list, then recalculating the total size of the new command list, and then de-padding at the end of that list. There's really no difference from an ordinary insert that operates in the middle of the command list, except for the offset.

**Deletion:**

```
aaabbcccc00000   # (0) original
aaacccc00000     # (1) removed load command b
aaacccc0000000   # (2) inserted zero bytes after last load command (pad)
```

**Insertion:**

```
aaabbcccc00000   # (0) original
aaaddbbcccc00000 # (1) inserted load command d before b
aaaddbbcccc000   # (2) removed zero bytes after last load command (de-pad)
```

**Addition (insertion at the end):**

```
aaabbcccc00000   # (0) original
aaabbccccdd00000 # (1) inserted load command d after c (before trailing zero bytes)
aaabbccccdd000   # (3) removed zero bytes after last load command (de-pad)
```

Does this make sense? If you still see a problem, which of the steps is problematic?

woodruffw: > Does this make sense?

Yep, this makes perfect sense. I was getting caught up on my older (replacement) implementation - your diagram clarified things :+1:

lib/macho/macho_file.rb
```diff
@@ -184,12 +186,43 @@ def add_command(lc)
         raise HeaderPadError.new(@filename)
       end
 
-      # update the raw data to reflect the new command
+      # update Mach-O header fields to account for deleted load command
```
UniqMartin: “[…] to account for **added** load command”.

lib/macho/macho_file.rb
```diff
@@ -172,10 +170,14 @@ def command(name)
...
     # @param lc [MachO::LoadCommand] the load command being added
+    # @param repopulate [Boolean] whether or not to repopulate the instance fields
     # @return [void]
     # @raise [MachO::HeaderPadError] if the new command exceeds the header pad buffer
     # @note This is public, but methods like {#add_rpath} should be preferred.
-    def add_command(lc)
+    #  Setting `repopulate` to false **will leave the instance in an
+    #  inconsistent state** unless {#populate_fields} is called **immediately**
+    #  afterwards.
+    def add_command(lc, repopulate = true)
```
UniqMartin: Making `repopulate` a simple boolean argument is convenient, but can be difficult to understand on the call site, where it's not clear what `false` could be meaning without consulting the documentation. One solution to this would be this method signature:

```ruby
def add_command(lc, options = {})
```

With the check then modified to:

```ruby
populate_fields if options.fetch(:repopulate, true)
```

And the call site now looking something like this:

```ruby
add_command(lc, :repopulate => false)
```

woodruffw: > Making repopulate a simple boolean argument is convenient, but can be difficult to understand on the call site

I was thinking the same thing. Having it as an option is certainly more obvious, and gives us future flexibility :+1:

lib/macho/macho_file.rb
woodruffw: Ended up going with a public-facing `#replace_command` backed internally by `#insert_command`, since we have to perform a header pad check even before deletion occurs.

UniqMartin: Sounds reasonable, though I'm not entirely sure why, as deleting a command should never cause the load commands to not fit into the available space?! Anyway, looking into the code now, so maybe this question will answer itself.

lib/macho/macho_file.rb
```diff
@@ -170,26 +168,68 @@ def command(name)
...
-    # @note This is public, but methods like {#add_rpath} should be preferred.
-    def add_command(lc)
+    # @note This is public, but methods like {#dylib_id=} should be preferred.
+    def replace_command(old_lc, new_lc)
       context = LoadCommand::SerializationContext.context_for(self)
-      cmd_raw = lc.serialize(context)
+      cmd_raw = new_lc.serialize(context)
       new_sizeofcmds = sizeofcmds + cmd_raw.bytesize
-
       if header.class.bytesize + new_sizeofcmds > low_fileoff
```
UniqMartin: Am I missing something or are you not taking into account the space freed up by deleting `old_lc`?

lib/macho/macho_file.rb
```diff
@@ -170,26 +168,68 @@ def command(name)
...
+    # @param lc [MachO::LoadCommand] the load command being added
+    # @param options [Hash]
+    # @option options [Boolean] :repopulate (true) whether or not to repopulate
+    #  the instance fields
+    # @return [void]
+    # @note This is public, but methods like {#add_rpath} should be preferred.
+    #  Setting `repopulate` to false **will leave the instance in an
+    #  inconsistent state** unless {#populate_fields} is called **immediately**
+    #  afterwards.
+    def add_command(lc, options = {})
```
UniqMartin: I wonder if this should be documenting the `MachO::HeaderPadError` exception possibly thrown via `insert_command` …

lib/macho/macho_file.rb
```diff
@@ -519,87 +549,30 @@ def set_sizeofcmds(size)
...
-      old_pad = old_padded_size - (old_prepad)
-      new_pad = new_padded_size - (new_prepad)
-
-      # pad the old and new strings with null bytes to meet command bounds
-      old_str << "\x00" * old_pad
-      new_str << "\x00" * new_pad
-
-      # calculate the new sizeofcmds in MH
-      new_sizeofcmds += new_padded_size - cmd.cmdsize
+    def insert_command(offset, lc, options = {})
```
UniqMartin: I'm missing a mention of the `MachO::HeaderPadError` exception in the documentation.

lib/macho/macho_file.rb
```diff
@@ -519,87 +549,30 @@ def set_sizeofcmds(size)
...
+      context = LoadCommand::SerializationContext.context_for(self)
+      cmd_raw = lc.serialize(context)
+      new_sizeofcmds = sizeofcmds + cmd_raw.bytesize
 
-      if new_sizeofcmds + header.class.bytesize > low_fileoff
+      if header.class.bytesize + new_sizeofcmds > low_fileoff
```
UniqMartin: I think it makes sense to do some basic checks on `offset`, namely whether it is between `0` (or `header.class.bytesize`?) and `low_fileoff` and whether `offset + size_of_lc` still stays below this upper limit. A more thorough check could also ensure `offset` isn't picked completely arbitrarily and causes a write in the middle of a load command, but I think it's also fine to trust the user to get this part right.

test/test_macho.rb
```diff
@@ -313,8 +321,15 @@ def test_change_install_name
 
       file.write(actual)
 
-      # compare actual and expected file hashes, to ensure file correctness
-      assert equal_sha1_hashes(actual, expected)
```
UniqMartin: No bit-for-bit equality here?

UniqMartin:
This has shaped up really nicely and I enjoy the flexibility and the comparatively simple code of `(add|delete|insert|replace)_command`! I'm excited to find out how we'll be using them in the future.

I left a few mostly minor comments, but overall we are very close to shipping this. :+1: :tada:

lib/macho/macho_file.rb
woodruffw: You're right, this is overly greedy. 

`new_sizeofcmds = sizeofcmds + cmd_raw.bytesize - old_cmd.cmdsize` should be correct, right?

lib/macho/macho_file.rb
woodruffw: We could, but my understanding of YARD is that you don't normally document an exception at a site unless that function throws it itself. Making an exception (no pun intended) here is reasonable though, since we don't expose `insert_command` as part of the public API.

lib/macho/macho_file.rb
woodruffw: Yep, fixing this.

lib/macho/macho_file.rb
woodruffw: Agreed, we definitely don't want users to be able to insert a load command in the middle of their header :smile:

test/test_macho.rb
woodruffw: Yeah, bit-for-bit equality is probably preferable to sha1 hashes (even if the chance of a hash collision is vanishingly small). Will fix :+1:

lib/macho/macho_file.rb
UniqMartin: Correct! This will become interesting once we get to the point where all of these are fully implemented and we try the code on some test files with barely enough or even insufficient available space.

lib/macho/macho_file.rb
woodruffw: Not because deleting a load command will set off the padding condition, but because we don't want an insertion to fail halfway through. If we let the deletion occur successfully (which it always will) and run into a failure during insertion, the resulting object will be left without that load command.

lib/macho/macho_file.rb
UniqMartin: I'll leave that up to you; both variants make sense to me and that's also why my original comment is fairly vague. Mainly wanted to see whether you thought about this and if not, provide a subtle reminder to do so. 😸

lib/macho/macho_file.rb
woodruffw: Actually, users should never hit this method directly anyways (since it's marked private both in the object and API), but doing sanity checks on `offset` is still a great idea.

test/test_macho.rb
UniqMartin: I think you misunderstood me. I'm perfectly fine with a hash comparison and it being treated equally with bit-for-bit equality. What I was referring to, was the fact that the hash comparison is being removed here.

woodruffw: Whoops! I must've missed it :sweat_smile:

lib/macho/macho_file.rb
UniqMartin: Yep, some basic sanity checks will be enough. I'm definitely not expecting you to perform an exhaustive and 100% bulletproof check on this method argument. 😸

lib/macho/macho_file.rb
woodruffw: I'm actually not 100% sure what to do here :confounded:. It would definitely be *nice* to warn users that `add_command` can fail during `insert_command`, but I've been following the practice of only documenting exceptions at their origin sites everywhere else in `ruby-macho`.

One solution to this might be to get rid of the `private` area of `MachOFile` entirely and then just `@see {#insert_command}`. This seems to be a fairly common practice, at least based on the ruby libraries I regularly use...

UniqMartin: I think exposing `insert_command` as an API (though as one very few people should be using directly) and then referencing it makes sense to me. To be honest, I haven't used YARD this much (probably less than you have) and I'm only dimly aware of the documentation practices of other Ruby library authors.

woodruffw: Agreed, this seems like the best of both worlds!
I'm going to add an extra `@note` to `insert_command` as well, just to make it clear that it's not safe to insert at any valid offset in the load command region.

test/test_macho.rb
UniqMartin: Ping on this code comment. I think all other comments in this PR have been addressed.

woodruffw: I added it above, right above the creation of `act` and `exp`. I just realized that it's missing from `test_change_rpath`, though...

UniqMartin: Ahem … `test_change_rpath` is the method these diff lines belong to and that's where I've been missing this. (And that's also why this isn't hidden yet in the PR view.) :wink:

woodruffw: Oh! That's what I get for only paying attention to the diff comment:

```diff
@@ -313,8 +321,15 @@ def test_change_install_name
```

I'll have this corrected tonight :smile:

UniqMartin: I don't blame you. It's quite evil of Git/GitHub to suggest the wrong context here …

woodruffw:
Anything else necessary on this PR before merging?

I forgot to follow up on https://github.com/Homebrew/ruby-macho/pull/45#discussion_r71095569, but I'll have that done tonight.

UniqMartin:
Apologies for the massive delay on this one (last week was a bit hectic)! Thanks for making some final tweaks to the documentation!

(I'll follow up on the two discussion issues you posted meanwhile as soon as possible.)

PR #49, opened 2016-07-28
Rpath duplication prevention/bugfix.

woodruffw:
Prevents `MachOFile#change_rpath` from creating duplicates, and fixes a variable in `MachOFile#delete_rpath`.

ref #41
cc @UniqMartin

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -16,6 +16,7 @@ module MachO
     attr_reader :header
 
     # @return [Array<MachO::LoadCommand>] an array of the file's load commands
+    # @note load commands are provided in order of ascending offset.
     attr_reader :load_commands
 
     # Creates a new MachOFile instance from a binary string.
@@ -363,13 +364,16 @@ module MachO
     # @param new_path [String] the new runtime path
     # @return [void]
     # @raise [MachO::RpathUnknownError] if no such old runtime path exists
+    # @raise [MachO::RpathExistsError] if the new runtime path already exists
     def change_rpath(old_path, new_path)
       old_lc = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
       raise RpathUnknownError.new(old_path) if old_lc.nil?
+      raise RpathExistsError.new(new_path) if rpaths.include?(new_path)
 
       new_lc = LoadCommand.create(:LC_RPATH, new_path)
 
-      replace_command(old_lc, new_lc)
+      delete_rpath(old_path)
+      insert_command(old_lc.view.offset, new_lc)
     end
 
     # Add the given runtime path to the Mach-O.
@@ -397,11 +401,7 @@ module MachO
     # @raise [MachO::RpathUnknownError] if no such runtime path exists
     def delete_rpath(path)
       rpath_cmds = command(:LC_RPATH).select { |r| r.path.to_s == path }
-      raise RpathUnknownError.new(old_path) if rpath_cmds.empty?
-
-      # sort_by! should be unnecessary since load commands are read in
-      # sequential order, but let's not rely on this behavior
-      rpath_cmds.sort_by! { |lc| lc.view.offset }
+      raise RpathUnknownError.new(path) if rpath_cmds.empty?
 
       # delete the commands in reverse order, offset descending. this
       # allows us to defer (expensive) field population until the very end
--- a/test/test_macho.rb
+++ b/test/test_macho.rb
@@ -46,6 +46,16 @@ class MachOFileTest < Minitest::Test
     end
   end
 
+  def test_load_commands_well_ordered
+    filenames = SINGLE_ARCHES.map { |a| fixture(a, "libhello.dylib") }
+
+    filenames.each do |fn|
+      file = MachO::MachOFile.new(fn)
+      offsets = file.load_commands.map { |lc| lc.view.offset }
+      assert_equal offsets.sort, offsets
+    end
+  end
+
   def test_mach_header
     filenames = SINGLE_ARCHES.map { |a| fixture(a, "libhello.dylib") }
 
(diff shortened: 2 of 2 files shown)


lib/macho/macho_file.rb
```diff
@@ -366,6 +366,7 @@ def rpaths
     def change_rpath(old_path, new_path)
       old_lc = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
       raise RpathUnknownError.new(old_path) if old_lc.nil?
+      raise RpathExistsError.new(new_path) if rpaths.include?(new_path)
```
UniqMartin: This should also be documented in the method documentation block.

lib/macho/macho_file.rb
```diff
@@ -397,7 +398,7 @@ def add_rpath(path)
     # @raise [MachO::RpathUnknownError] if no such runtime path exists
     def delete_rpath(path)
       rpath_cmds = command(:LC_RPATH).select { |r| r.path.to_s == path }
-      raise RpathUnknownError.new(old_path) if rpath_cmds.empty?
+      raise RpathUnknownError.new(path) if rpath_cmds.empty?
```
UniqMartin: Good catch!

UniqMartin:
The extra check you added to prevent introducing new duplicate RPATHs is good! :+1:

What you didn't change yet, but what I had mainly in mind, was (like for RPATH deletion) to make sure that the change also eliminates duplicates of the to-be-modified RPATH, e.g.:

```ruby
file.rpaths
# => ["/usr/lib", "/usr/lib", "/opt/homebrew/lib"]
file.change_rpath("/usr/lib", "/usr/local/lib")
file.rpaths
# => ["/usr/local/lib", "/opt/homebrew/lib"]
```

Unless I'm overlooking something, the result with the current implementation instead will be:

```ruby
file.rpaths
# => ["/usr/local/lib", "/usr/lib", "/opt/homebrew/lib"]
```

woodruffw:
> What you didn't change yet, but what I had mainly in mind, was (like for RPATH deletion) to make sure that the change also eliminates duplicates of the to-be-modified RPATH

Whoops, forgot to push it...

lib/macho/macho_file.rb
```diff
@@ -363,13 +363,16 @@ def rpaths
...
     def change_rpath(old_path, new_path)
       old_lc = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
       raise RpathUnknownError.new(old_path) if old_lc.nil?
+      raise RpathExistsError.new(new_path) if rpaths.include?(new_path)
 
       new_lc = LoadCommand.create(:LC_RPATH, new_path)
 
-      replace_command(old_lc, new_lc)
+      delete_rpath(old_path)
+      insert_command(old_lc.view.offset, new_lc)
```
UniqMartin: Clever! I was looking forward to some fiddling with `replace_command` and `delete_command`. :wink:

The only concern that I have, because we also haven't made this assumption in `delete_rpath`, is that the first match will be automatically the first one inside the Mach-O file (in terms of file offset).

woodruffw: That assumption is currently correct, but you're right that it's not a good one to make. I can eliminate it by copying a bit of the logic we do in `delete_rpath` - using `select` and `sort_by!` instead of `find`. Does that sound good to you?

UniqMartin: Yes, either that or we also start making this assumption in `delete_rpath`. The latter option seems also reasonable to me, but for me to accept it I'd like to see that documented (make the assumed sort order official) and enforced in tests (make sure the order stays the same if we end up refactoring some bits). There's of course an advantage in avoiding the sort if we know the array is always pre-sorted.

Pick your favorite option. 😸 Independently of the option (both are valid), I tend to prefer a defensive coding style for operations like this one. Otherwise we'll easily get corrupted Mach-O binaries if something unexpectedly or accidentally changes in the code base.

woodruffw: > The latter option seems also reasonable to me, but for me to accept it I'd like to see that documented (make the assumed sort order official) and enforced in tests (make sure the order stays the same if we end up refactoring some bits).

This sounds like the path of least resistance to me :+1:

test/helpers.rb
```diff
@@ -59,6 +59,10 @@ def filechecks(except = nil)
     checks
   end
 
+  def array_sorted?(array)
+    array.each_cons(2).all? { |i, j| (i <=> j) <= 0 }
```
UniqMartin: The chaining of `<=>` and `<=` is a bit redundant, isn't it? `i <= j` will have the exact same effect. (But read on in the next code comment before you address this.)

test/test_macho.rb
```diff
@@ -46,6 +46,16 @@ def test_load_commands
     end
   end
 
+  def test_load_commands_well_ordered
+    filenames = SINGLE_ARCHES.map { |a| fixture(a, "libhello.dylib") }
+
+    filenames.each do |fn|
+      file = MachO::MachOFile.new(fn)
+      offsets = file.load_commands.map(&:view).map(&:offset)
+      assert array_sorted?(offsets)
```
UniqMartin: I appreciate the effort to create a reusable predicate method for this check. However, I think this test code is not performance critical, thus I'd prefer to use something really simple like:

```ruby
assert_equal offsets.sort, offsets
```

Another advantage is that this by default results in a more useful message if the assertion fails.

test/test_macho.rb
```diff
@@ -46,6 +46,16 @@ def test_load_commands
     end
   end
 
+  def test_load_commands_well_ordered
+    filenames = SINGLE_ARCHES.map { |a| fixture(a, "libhello.dylib") }
+
+    filenames.each do |fn|
+      file = MachO::MachOFile.new(fn)
+      offsets = file.load_commands.map(&:view).map(&:offset)
```
UniqMartin: I don't have very strong feelings regarding this (thus feel free to ignore and keep as-is), but I'm leaning towards the following instead of chaining multiple `map` calls:

```ruby
offsets = file.load_commands.map { |lc| lc.view.offset }
```

test/test_macho.rb
woodruffw: Works for me!

UniqMartin:
Awesome; thanks for steadily grinding through the remaining issues and known bugs! :tada:

Issue #51, opened 2016-07-29
Changing install name can break the binary


UniqMartin:
`MachOFile#change_install_name` corrupts load commands in the sense that it doesn't preserve the type of the command when changing the name. This will happen rarely, but wen it does, it usually changes semantics and thus breaks the binary. (Sorry I missed this earlier in review.)

Steps to reproduce:

```console
$ echo "void foo() {}" > lib.c
$ clang -dynamiclib -o libtest.dylib -Wl,-reexport-lz lib.c
$ otool -l libtest.dylib > otool-old.log
$ ruby -Ilib -rmacho -e 'MachO::Tools.change_install_name("libtest.dylib", "/usr/lib/libz.1.dylib", "does_not_matter")'
$ otool -l libtest.dylib > otool-new.log
```

Notice how `LC_REEXPORT_DYLIB` has changed to `LC_LOAD_DYLIB`:

```diff
diff --git 1/otool-old.log 2/otool-new.log
index 55435787..62636422 100644
--- 1/otool-old.log
+++ 2/otool-new.log
@@ -109,9 +109,9 @@ Load command 8
   cmdsize 16
   version 0.0
 Load command 9
-          cmd LC_REEXPORT_DYLIB
-      cmdsize 48
-         name /usr/lib/libz.1.dylib (offset 24)
+          cmd LC_LOAD_DYLIB
+      cmdsize 40
+         name does_not_matter (offset 24)
    time stamp 2 Thu Jan  1 01:00:02 1970
       current version 1.2.5
 compatibility version 1.0.0
```

cc @woodruffw

woodruffw:
Luckily, this is a single-line fix in `MachOFile#change_install_name`:

```ruby
LoadCommand.create(:LC_LOAD_DYLIB, ...)
```

to

```ruby
LoadCommand.create(old_lc.type, ...)
```

PR will be ready in a few minutes :smile:

PR #53, opened 2016-07-29
macho_file: Preserve a dylib's type when changing it.

woodruffw:
Fixes a bug where all changed install names were turned into
LC_LOAD_DYLIB commands.

ref #51 
cc @UniqMartin

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -342,7 +342,7 @@ module MachO
       old_lc = dylib_load_commands.find { |d| d.name.to_s == old_name }
       raise DylibUnknownError.new(old_name) if old_lc.nil?
 
-      new_lc = LoadCommand.create(:LC_LOAD_DYLIB, new_name,
+      new_lc = LoadCommand.create(old_lc.type, new_name,
         old_lc.timestamp, old_lc.current_version, old_lc.compatibility_version)
 
       replace_command(old_lc, new_lc)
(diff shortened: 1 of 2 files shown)


UniqMartin:
Could you add a wee little test that would have revealed this mistake?

You can also append a line like `Closes #<issue-number>.` to your commit message (here and in similar cases; typically separated by an empty line from the rest of the commit message) to auto-close the related issue once the commit is merged to `master`. (This seems to be good practice on GitHub and leaves a nice cross reference in the Git history.)

woodruffw:
> Could you add a wee little test that would have revealed this mistake?

Sure!

> You can also append a line like Closes #<issue-number>. to your commit message (here and in similar cases; typically separated by an empty line from the rest of the commit message) to auto-close the related issue once the commit is merged to master.

No problem, I'll do this from now on :+1:

UniqMartin:
Thanks for the quick turnaround on this! :tada:

> Sure!

This looks good!

> No problem, I'll do this from now on 👍

Note that this should be the issue number (#51), not the number of this PR (#53). The latter will auto-close anyway. (I quickly fixed up the commit message myself before merging.)

woodruffw:
> Note that this should be the issue number (#51), not the number of this PR (#53). The latter will auto-close anyway. (I quickly fixed up the commit message myself before merging.)

Oops :sweat_smile:

Issue #52, opened 2016-07-29
LCStr: Incorrect handling of null-terminated strings


UniqMartin:
The `alot` issue has also revealed another interesting, though completely unrelated, issue. Here's what I did with the `alot-python` binary (can provide it in full if it helps) after copying it from `Cellar/alot/0.3.7/libexec/bin/python`:

```console
$ ruby -Ilib -rmacho -e 'puts MachO::Tools.dylibs("alot-python").sort' > dylibs-rb.log
$ otool -L alot-python | tail -n +2 | awk '{print $1}' | sort > dylibs-cc.log
```

The resulting diff:

```diff
diff --git 1/dylibs-cc.log 2/dylibs-rb.log
index 6286cda5..a0b685cc 100644
--- 1/dylibs-cc.log
+++ 2/dylibs-rb.log
@@ -1,3 +1,3 @@
 /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation
 /usr/lib/libSystem.B.dylib
-@executable_path/../.Python
+@executable_path/../.Pythonython.framework/Versions/2.7/Python
```

The relevant section from the Mach-O file (the problematic load command starts at 0x13dc):

```console
$ hexdump -vC alot-python
[…snip…]
000013c0  00 5e 00 00 28 00 00 80  18 00 00 00 07 0e 00 00  |.^..(...........|
000013d0  00 00 00 00 00 00 00 00  00 00 00 00 0c 00 00 00  |................|
000013e0  58 00 00 00 18 00 00 00  02 00 00 00 0a 07 02 00  |X...............|
000013f0  00 07 02 00 40 65 78 65  63 75 74 61 62 6c 65 5f  |....@executable_|
00001400  70 61 74 68 2f 2e 2e 2f  2e 50 79 74 68 6f 6e 00  |path/../.Python.|
00001410  79 74 68 6f 6e 2e 66 72  61 6d 65 77 6f 72 6b 2f  |ython.framework/|
00001420  56 65 72 73 69 6f 6e 73  2f 32 2e 37 2f 50 79 74  |Versions/2.7/Pyt|
00001430  68 6f 6e 00 0c 00 00 00  34 00 00 00 18 00 00 00  |hon.....4.......|
00001440  02 00 00 00 01 0a ca 04  00 00 01 00 2f 75 73 72  |............/usr|
[...]

PR #54, opened 2016-07-29
load_commands: Grab only the first null terminated string.

woodruffw:
Corrects behavior when the (normally null) padding of a load command
contains garbage data instead.

ref #52 
cc @UniqMartin

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -258,6 +258,7 @@ module MachO
       # @param lc [MachO::LoadCommand] the load command
       # @param lc_str [Fixnum, String] the offset to the beginning of the string,
       #  or the string itself if not being initialized with a view.
+      # @raise [MachO::LCStrMalformedError] if the string is malformed
       # @todo devise a solution such that the `lc_str` parameter is not
       #  interpreted differently depending on `lc.view`. The current behavior
       #  is a hack to allow viewless load command creation.
@@ -268,7 +269,9 @@ module MachO
         if view
           lc_str_abs = view.offset + lc_str
           lc_end = view.offset + lc.cmdsize - 1
-          @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
+          raw_string = view.raw_data.slice(lc_str_abs..lc_end)
+          @string, null_byte, _ = raw_string.partition("\x00")
+          raise LCStrMalformedError.new(lc) if null_byte.empty?
           @string_offset = lc_str
         else
           @string = lc_str
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -106,6 +106,14 @@ module MachO
     end
   end
 
+  # Raised when a load command string is malformed in some way.
+  class LCStrMalformedError < MachOError
+    # @param lc [MachO::LoadCommand] the load command containing the string
+    def initialize(lc)
+      super "Load command #{lc.type} at offset #{lc.view.offset} contains a malformed string"
+    end
+  end
+
   # Raised when a change at an offset is not valid.
   class OffsetInsertionError < MachOError
     # @param offset [Fixnum] the invalid offset


lib/macho/load_commands.rb
```diff
@@ -268,7 +268,8 @@ def initialize(lc, lc_str)
         if view
           lc_str_abs = view.offset + lc_str
           lc_end = view.offset + lc.cmdsize - 1
-          @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
+          raw_string = view.raw_data.slice(lc_str_abs..lc_end)
+          @string = raw_string.split("\x00").first
```
UniqMartin: Note that `String#partition` or `String#rpartition` is a more efficient alternative to `String#split` if we're only interested in the first/last delimited chunk. It's also less error-prone if we ever come across an empty string (valid, but consists entirely of null bytes). Try `"\x00\x00\x00".split("\x00")` to see what I mean.

In this case I would prefer a solution that makes sure the terminating `\x00` is always there, to make sure we're actually dealing with a null-terminated string. This could work (and was my first idea):

```ruby
null_index = raw_string.index("\x00")
raise "An appropriate exception" if null_index.nil?
@string = raw_string[0...null_index]
```

Using `String#partition` turns out to be more succinct (and still allows for validation):

```ruby
@string, null_byte, _ = raw_string.partition("\x00")
raise "An appropriate exception" if null_byte != "\x00"
```

Another simple alternative (that I'm less fond of because it doesn't validate):

```ruby
raw_string.unpack("Z*")
```

(Sorry for the potentially confusingly long comment. I prefer the `String#partition` variant, but I figured you might benefit from seeing all the other stuff that popped into my head while thinking about this.)

woodruffw: I'm leaning towards `String#partition` as well.

How about `LCStrMissingError`, in the style of `DylibIdMissingError`?

UniqMartin: I think `MalformedLCStrError` or `LCStrMalformedError` better captures the nature of the error.

woodruffw: `LCStrMalformedError` is great!

woodruffw: I made the condition `@string.empty? || null_byte.empty?`, since we probably also want to prevent the user from doing something like `change_install_name(old, "\x00")`.

lib/macho/load_commands.rb
```diff
@@ -268,7 +269,9 @@ def initialize(lc, lc_str)
         if view
           lc_str_abs = view.offset + lc_str
           lc_end = view.offset + lc.cmdsize - 1
-          @string = view.raw_data.slice(lc_str_abs..lc_end).delete("\x00")
+          raw_string = view.raw_data.slice(lc_str_abs..lc_end)
+          @string, null_byte, _ = raw_string.partition("\x00")
+          raise LCStrMalformedError.new(lc) if @string.empty? || null_byte.empty?
```
UniqMartin: > I made the condition `@string.empty? || null_byte.empty?`, since we probably also want to prevent the user from doing something like `change_install_name(old, "\x00")`.

That's a good intention, but I think we have to tackle validation of user input separately (and more thoroughly, if we want to go down that route). Let's focus on the bug fix here!

That's also not the code path being taken if the user invokes `LoadCommand.create` (or any of the helper methods that make use of it). That case is handled in the `else` branch below. Here we're dealing with data coming directly from the binary.

test/test_macho.rb
```diff
@@ -46,6 +46,15 @@ def test_load_commands
     end
   end
 
+  def test_load_command_string_malformed
+    filename = fixture(:i386, "hello.bin")
+    file = MachO::MachOFile.new(filename)
+
+    assert_raises MachO::LCStrMalformedError do
+      file.change_install_name(file.linked_dylibs.first, "\x00bad")
```
UniqMartin: I was confused for a moment why this indeed raises the exception, but it does so after writing the malformed `LCStr` and when re-reading all load commands. (I think we should be preventing this problematic user input earlier, but as stated in my other comment, not as part of this PR.)

lib/macho/load_commands.rb
woodruffw: > I think we have to tackle validation of user input separately

Agreed, I'll remove this.

test/test_macho.rb
woodruffw: Yep, that's the behavior that I was counting on. Now that we're removing this, this test probably won't be of much user either...

lib/macho/load_commands.rb
UniqMartin: Thanks for understanding! I appreciate this, William. :bow:

UniqMartin:
Thanks for the fix!

Issue #50, opened 2016-07-29
Treatment of discrepancies in Mach-O slices of fat binaries


UniqMartin:
As has been raised in https://github.com/Homebrew/brew/pull/592#issuecomment-235796041 and subsequent comments, test bots running `ruby-macho` fail to bottle `alot` from PR Homebrew/homebrew-core#1663:

```console
$ brew bottle --verbose --debug --json alot
/opt/brewery/dummy/Library/Homebrew/brew.rb (Formulary::TapLoader): loading /opt/brewery/dummy/Library/Taps/homebrew/homebrew-core/Formula/alot.rb
==> Determining alot bottle revision...
==> Bottling alot-0.3.7.el_capitan.bottle.tar.gz...
Changing install name in /opt/brewery/dummy/Cellar/alot/0.3.7/libexec/lib/python2.7/site-packages/gpgme/_gpgme.so
  from /opt/brewery/dummy/opt/gpgme/lib/libgpgme.11.dylib
    to @@HOMEBREW_PREFIX@@/opt/gpgme/lib/libgpgme.11.dylib
Error: No such dylib name: /opt/brewery/dummy/opt/gpgme/lib/libgpgme.11.dylib
/opt/brewery/dummy/Library/Homebrew/vendor/macho/macho/macho_file.rb:240:in `change_install_name'
/opt/brewery/dummy/Library/Homebrew/vendor/macho/macho/fat_file.rb:171:in `block in change_install_name'
/opt/brewery/dummy/Library/Homebrew/vendor/macho/macho/fat_file.rb:170:in `each'
/opt/brewery/dummy/Library/Homebrew/vendor/macho/macho/fat_file.rb:170:in `change_install_name'
/opt/brewery/dummy/Library/Homebrew/vendor/macho/macho/tools.rb:33:in `change_install_name'
/opt/brewery/dummy/Library/Homebrew/os/mac/ruby_keg.rb:13:in `change_install_name'
[…snip…]
```

The reason turns out to be that the problematic file is a fat binary, but not all of its slices link to the mentioned dylib (most likely because the dylib in question only has a single architecture):

```console
$ otool -arch all -L /opt/brewery/dummy/Cellar/alot/0.3.7/libexec/lib/python2.7/site-packages/gpgme/_gpgme.so
/opt/brewery/dummy/Cellar/alot/0.3.7/libexec/lib/python2.7/site-packages/gpgme/_gpgme.so (architecture i386):
[...]

PR #55, opened 2016-07-29
Allow suppression of exceptions with strict: false

woodruffw:
Resolves the problem of valid discrepancies between individual Mach-Os in a fat
file causing exceptions during modifications.

ref #50 
cc @UniqMartin

### DIFF
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -3,6 +3,26 @@ module MachO
   class MachOError < RuntimeError
   end
 
+  # Raised when a Mach-O file modification fails.
+  class ModificationError < MachOError
+  end
+
+  # Raised when a Mach-O file modification fails but can be recovered when
+  # operating on multiple Mach-O slices of a fat binary in non-strict mode.
+  class RecoverableModificationError < ModificationError
+    # @return [Fixnum, nil] The index of the Mach-O slice of a fat binary for
+    #   which modification failed or `nil` if not a fat binary. This is used to
+    #   make the error message more useful.
+    attr_accessor :macho_slice
+
+    # @return [String] The exception message.
+    def to_s
+      s = super.to_s
+      s = "While modifying Mach-O slice #{@macho_slice}: #{s}" if @macho_slice
+      s
+    end
+  end
+
   # Raised when a file is not a Mach-O.
   class NotAMachOError < MachOError
     # @param error [String] the error in question
@@ -115,7 +135,7 @@ module MachO
   end
 
   # Raised when a change at an offset is not valid.
-  class OffsetInsertionError < MachOError
+  class OffsetInsertionError < ModificationError
     # @param offset [Fixnum] the invalid offset
     def initialize(offset)
       super "Insertion at offset #{offset} is not valid"
@@ -123,7 +143,7 @@ module MachO
   end
 
   # Raised when load commands are too large to fit in the current file.
-  class HeaderPadError < MachOError
+  class HeaderPadError < ModificationError
     # @param filename [String] the filename
     def initialize(filename)
       super "Updated load commands do not fit in the header of " +
@@ -133,7 +153,7 @@ module MachO
   end
 
   # Raised when attempting to change a dylib name that doesn't exist.
-  class DylibUnknownError < MachOError
+  class DylibUnknownError < RecoverableModificationError
     # @param dylib [String] the unknown shared library name
     def initialize(dylib)
       super "No such dylib name: #{dylib}"
@@ -141,14 +161,14 @@ module MachO
   end
 
   # Raised when a dylib is missing an ID
-  class DylibIdMissingError < MachOError
+  class DylibIdMissingError < RecoverableModificationError
     def initialize
       super "Dylib is missing a dylib ID"
     end
   end
 
   # Raised when attempting to change an rpath that doesn't exist.
-  class RpathUnknownError < MachOError
+  class RpathUnknownError < RecoverableModificationError
     # @param path [String] the unknown runtime path
     def initialize(path)
       super "No such runtime path: #{path}"
@@ -156,7 +176,7 @@ module MachO
   end
 
   # Raised when attempting to add an rpath that already exists.
-  class RpathExistsError < MachOError
+  class RpathExistsError < RecoverableModificationError
     # @param path [String] the extant path
     def initialize(path)
       super "#{path} already exists"
--- a/lib/macho/fat_file.rb
+++ b/lib/macho/fat_file.rb
@@ -133,12 +133,15 @@ module MachO
 
     # Changes the file's dylib ID to `new_id`. If the file is not a dylib, does nothing.
     # @example
-    #  file.dylib_id = 'libFoo.dylib'
+    #  file.change_dylib_id('libFoo.dylib')
     # @param new_id [String] the new dylib ID
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) if true, fail if one slice fails.
+    #  if false, fail only if all slices fail.
     # @return [void]
     # @raise [ArgumentError] if `new_id` is not a String
     # @see MachO::MachOFile#linked_dylibs
-    def dylib_id=(new_id)
+    def change_dylib_id(new_id, options = {})
       if !new_id.is_a?(String)
         raise ArgumentError.new("argument must be a String")
       end
@@ -147,12 +150,14 @@ module MachO
         return nil
       end
 
-      machos.each do |macho|
-        macho.dylib_id = new_id
+      each_macho(options) do |macho|
+        macho.change_dylib_id(new_id, options)
       end
 
       synchronize_raw_data
     end
+
+    alias :dylib_id= :change_dylib_id
 
     # All shared libraries linked to the file's Mach-Os.
     # @return [Array<String>] an array of all shared libraries
@@ -170,11 +175,14 @@ module MachO
     #  file.change_install_name('/usr/lib/libFoo.dylib', '/usr/lib/libBar.dylib')
     # @param old_name [String] the shared library name being changed
     # @param new_name [String] the new name
-    # @todo incomplete
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) if true, fail if one slice fails.
+    #  if false, fail only if all slices fail.
+    # @return [void]
     # @see MachO::MachOFile#change_install_name
-    def change_install_name(old_name, new_name)
-      machos.each do |macho|
-        macho.change_install_name(old_name, new_name)
+    def change_install_name(old_name, new_name, options = {})
+      each_macho(options) do |macho|
+        macho.change_install_name(old_name, new_name, options)
       end
 
       synchronize_raw_data
@@ -193,11 +201,14 @@ module MachO
     # Change the runtime path `old_path` to `new_path` in the file's Mach-Os.
     # @param old_path [String] the old runtime path
     # @param new_path [String] the new runtime path
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) if true, fail if one slice fails.
+    #  if false, fail only if all slices fail.
     # @return [void]
     # @see MachO::MachOFile#change_rpath
-    def change_rpath(old_path, new_path)
-      machos.each do |macho|
-        macho.change_rpath(old_path, new_path)
+    def change_rpath(old_path, new_path, options = {})
+      each_macho(options) do |macho|
+        macho.change_rpath(old_path, new_path, options)
       end
 
       synchronize_raw_data
@@ -205,11 +216,14 @@ module MachO
 
     # Add the given runtime path to the file's Mach-Os.
     # @param path [String] the new runtime path
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) if true, fail if one slice fails.
+    #  if false, fail only if all slices fail.
     # @return [void]
     # @see MachO::MachOFile#add_rpath
-    def add_rpath(path)
-      machos.each do |macho|
-        macho.add_rpath(path)
+    def add_rpath(path, options = {})
+      each_macho(options) do |macho|
+        macho.add_rpath(path, options)
       end
 
       synchronize_raw_data
@@ -217,11 +231,14 @@ module MachO
 
     # Delete the given runtime path from the file's Mach-Os.
     # @param path [String] the runtime path to delete
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) if true, fail if one slice fails.
+    #  if false, fail only if all slices fail.
     # @return void
     # @see MachO::MachOFile#delete_rpath
-    def delete_rpath(path)
-      machos.each do |macho|
-        macho.delete_rpath(path)
+    def delete_rpath(path, options = {})
+      each_macho(options) do |macho|
+        macho.delete_rpath(path, options)
       end
 
       synchronize_raw_data
@@ -310,6 +327,34 @@ module MachO
       end
 
       machos
+    end
+
+    # Yield each Mach-O object in the file, rescuing and accumulating errors.
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) whether or not to fail loudly
+    #  with an exception if at least one Mach-O raises an exception. If false,
+    #  only raises an exception if *all* Mach-Os raise exceptions.
+    # @raise [MachO::RecoverableModificationError] under the conditions of
+    #  the `:strict` option above.
+    # @api private
+    def each_macho(options = {})
+      strict = options.fetch(:strict, true)
+      errors = []
+
+      machos.each_with_index do |macho, index|
+        begin
+          yield macho
+        rescue RecoverableModificationError => error
+          error.macho_slice = index
+
+          # Strict mode: Immediately re-raise. Otherwise: Retain, check later.
+          raise error if strict
+          errors << error
+        end
+      end
+
+      # Non-strict mode: Raise first error if *all* Mach-O slices failed.
+      raise errors.first if errors.size == machos.size
     end
 
     # Synchronize the raw file data with each internal Mach-O object.
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -305,11 +305,12 @@ module MachO
 
     # Changes the Mach-O's dylib ID to `new_id`. Does nothing if not a dylib.
     # @example
-    #  file.dylib_id = "libFoo.dylib"
+    #  file.change_dylib_id("libFoo.dylib")
     # @param new_id [String] the dylib's new ID
+    # @param options [Hash]
     # @return [void]
     # @raise [ArgumentError] if `new_id` is not a String
-    def dylib_id=(new_id)
+    def change_dylib_id(new_id, options = {})
       raise ArgumentError.new("new ID must be a String") unless new_id.is_a?(String)
       return unless dylib?
 
@@ -321,6 +322,8 @@ module MachO
 
       replace_command(old_lc, new_lc)
     end
+
+    alias :dylib_id= :change_dylib_id
 
     # All shared libraries linked to the Mach-O.
     # @return [Array<String>] an array of all shared libraries
@@ -337,11 +340,12 @@ module MachO
     #  file.change_install_name("/usr/lib/libWhatever.dylib", "/usr/local/lib/libWhatever2.dylib")
     # @param old_name [String] the shared library's old name
     # @param new_name [String] the shared library's new name
+    # @param options [Hash]
     # @return [void]
     # @raise [MachO::DylibUnknownError] if no shared library has the old name
-    def change_install_name(old_name, new_name)
+    def change_install_name(old_name, new_name, options = {})
       old_lc = dylib_load_commands.find { |d| d.name.to_s == old_name }
-      raise DylibUnknownError.new(old_name) if old_lc.nil?
+      raise DylibUnknownError.new(old_name) unless old_lc
 
       new_lc = LoadCommand.create(old_lc.type, new_name,
         old_lc.timestamp, old_lc.current_version, old_lc.compatibility_version)
@@ -362,12 +366,13 @@ module MachO
     #  file.change_rpath("/usr/lib", "/usr/local/lib")
     # @param old_path [String] the old runtime path
     # @param new_path [String] the new runtime path
+    # @param options [Hash]
     # @return [void]
     # @raise [MachO::RpathUnknownError] if no such old runtime path exists
     # @raise [MachO::RpathExistsError] if the new runtime path already exists
-    def change_rpath(old_path, new_path)
+    def change_rpath(old_path, new_path, options = {})
       old_lc = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
-      raise RpathUnknownError.new(old_path) if old_lc.nil?
+      raise RpathUnknownError.new(old_path) unless old_lc
       raise RpathExistsError.new(new_path) if rpaths.include?(new_path)
 
       new_lc = LoadCommand.create(:LC_RPATH, new_path)
@@ -382,9 +387,10 @@ module MachO
     #  file.add_rpath("/usr/lib")
     #  file.rpaths # => ["/lib", "/usr/lib"]
     # @param path [String] the new runtime path
+    # @param options [Hash]
     # @return [void]
     # @raise [MachO::RpathExistsError] if the runtime path already exists
-    def add_rpath(path)
+    def add_rpath(path, options = {})
       raise RpathExistsError.new(path) if rpaths.include?(path)
 
       rpath_cmd = LoadCommand.create(:LC_RPATH, path)
@@ -397,9 +403,10 @@ module MachO
     #  file.delete_rpath("/lib")
     #  file.rpaths # => []
     # @param path [String] the runtime path to delete
+    # @param options [Hash]
     # @return void
     # @raise [MachO::RpathUnknownError] if no such runtime path exists
-    def delete_rpath(path)
+    def delete_rpath(path, options = {})
       rpath_cmds = command(:LC_RPATH).select { |r| r.path.to_s == path }
       raise RpathUnknownError.new(path) if rpath_cmds.empty?
 
--- a/lib/macho/tools.rb
+++ b/lib/macho/tools.rb
@@ -12,11 +12,14 @@ module MachO
     # Changes the dylib ID of a Mach-O or Fat binary, overwriting the source file.
     # @param filename [String] the Mach-O or Fat binary being modified
     # @param new_id [String] the new dylib ID for the binary
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) whether or not to fail loudly
+    #  with an exception if the change cannot be performed
     # @return [void]
-    def self.change_dylib_id(filename, new_id)
+    def self.change_dylib_id(filename, new_id, options = {})
       file = MachO.open(filename)
 
-      file.dylib_id = new_id
+      file.change_dylib_id(new_id, options)
       file.write!
     end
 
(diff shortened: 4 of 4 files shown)


woodruffw:
Once this is in, I'll push a release and open a brew PR to update the vendored version/calls to modification methods.

UniqMartin:
I think we might have a misunderstanding what the strict vs. non-strict behavior should be. Or maybe I'm misreading your code. I'm assuming a fat file with more than one Mach-O slice in the following, but thin files (regular Mach-O files) can be treated as if they were fat files with a single slice. Here's what I had in mind:

1. If the change succeeds for all slices: success in both strict and non-strict mode.
2. If the change succeeds in some slices, but fails in others: failure in strict mode, but success in non-strict mode. (👈 **This is the change I intended.**)
3. If the change fails for all slices: failure in both strict and non-strict mode.

From reading the diff in this PR I got the impression you're also treating case (3) as success in non-strict mode, which is not something I'm willing to allow, as this would open the floodgates and will silently ignore too many problems.

I think the easiest way to achieve what I had in mind (aside from adding `options = {}` everywhere) would be to modify the methods in `FatFile` such that they catch exceptions for individual slices, count how many slices were modified successfully, and then re-raise or swallow the errors in accordance with the above list. Does this make sense? (I'm happy to add more clarifying words and/or code snippets.)

> Once this is in, I'll push a release and open a brew PR to update the vendored version/calls to modification methods.

It's not strictly necessary, but would be nice to have #49 in the release, too. Otherwise: Agreed!

woodruffw:
> From reading the diff in this PR I got the impression you're also treating case (3) as success in non-strict mode, which is not something I'm willing to allow, as this would open the floodgates and will silently ignore too many problems.

You're reading correctly, setting `strict: false` will currently let all slices fail silently, and this probably isn't good behavior.

> I think the easiest way to achieve what I had in mind (aside from adding options = {} everywhere) would be to modify the methods in FatFile such that they catch exceptions for individual slices, count how many slices were modified successfully, and then re-raise or swallow the errors in accordance with the above list.

This makes sense, although I don't generally like using exceptions as control flow. I think it might be easier (and faster) to accomplish the same thing with something like this at the end of each `FatFile` modification method:

```ruby
[...]

successful = machos.count { |m| m.dylib_id == new_id }
raise AnErrorExpressiveOfCompleteFailure if successful == 0

synchronize_raw_data
```

Where `AnErrorExpressiveOfCompleteFailure` is a new exception that distinguishes single-point failures from a complete failure to perform the modification on any slice. Thoughts?

> It's not strictly necessary, but would be nice to have #49 in the release, too.

Sounds good to me!

woodruffw:
@UniqMartin Any thoughts on my counterproposal for the case when all slices fail? I'd like to push this PR along to keep us on track with the schedule :smile:

UniqMartin:
Apologies for the delay! Since we're on the same page regarding the desired behavior, please next time just move ahead and make the necessary adjustments and implement them how you think it should be done, unless doing so would mean an excessive amount of work. It's always easier to discuss actual code and it's quite possible that you'll either notice design flaws in your idea while implementing it or the finished code turns out being superior to what I had in mind.

In any case and unless you already have done so, please use the time until you hear back from me to construct tests that check for the desired behavior, e.g. for this PR fat binaries with inconsistent Mach-O slices and test code that works with them.

> This makes sense, although I don't generally like using exceptions as control flow.

In general, I tend to agree here. The idea that was backing this suggestion would be that this would allow for richer error handling, where the exception raised in the top-level modification method that ensures the modification could be applied to any/all slices could more easily point out what went wrong and which Mach-O slice was responsible for this.

> I think it might be easier (and ~~faster~~ more efficient) to accomplish the same thing with something like this at the end of each `FatFile` modification method: […] Where `AnErrorExpressiveOfCompleteFailure` is a new exception that distinguishes single-point failures from a complete failure to perform the modification on any slice. Thoughts?

I think this idea works fairly well for the presented case of `LC_ID_DYLIB` but will quickly become more complex for the other load commands and unnecessarily so, since the called method that operates on a single Mach-O slice already does all the necessary checks.

woodruffw:
> I think this idea works fairly well for the presented case of LC_ID_DYLIB but will quickly become more complex for the other load commands and unnecessarily so, since the called method that operates on a single Mach-O slice already does all the necessary checks.

Yeah, that's true. Doing this by filtering exceptions sounds good to me with that in mind.

Do you have a particular structure in mind? Catching exceptions and counting them with an accumulator is what immediately comes to mind, but that doesn't feel very ruby-esque.



> In any case and unless you already have done so, please use the time until you hear back from me to construct tests that check for the desired behavior, e.g. for this PR fat binaries with inconsistent Mach-O slices and test code that works with them.

I'll get on this. I think there's a `cctools` utility for smashing binaries together (i.e., the opposite of `lipo(1)`), so we can probably leverage that to make some fat binaries with inconsistent slices.

woodruffw:
> i.e., the opposite of `lipo(1)`

Never mind, looks like `lipo(1)` does this directly with `-create` :smile:

UniqMartin:
> Do you have a particular structure in mind? Catching exceptions and counting them with an accumulator is what immediately comes to mind, but that doesn't feel very ruby-esque.

Roughly, this is what I had in mind:

```ruby
errors = []
machos.each_with_index do |macho, index|
  begin
    macho.‹operation›
  rescue AppropriateBaseClass => e
    errors << [index, e]
  end
end

strict = options.fetch(:strict, false)
failed = strict ? !errors.empty? : errors.size == machos.size
if failed
  macho_index, cause = errors.first
  raise MachOSliceModificationError.new(strict, macho_index, cause)
end
```

Here, `AppropriateBaseClass` could potentially be `ArgumentError` and `DylibUnknownError` and similar exception classes could become children of `ArgumentError`. (I haven't fully fleshed out this thought, but it would be nice to have a distinction between exception classes that are due to problems with the binary that we're reading and exceptions due to inappropriate use of our API, because I only want the latter to be rescued and possibly ignored if in non-strict mode.)

It obviously doesn't make sense to repeat all this in every one of the methods, but the logic could be easily encapsulated, such that, e.g., `FatFile#delete_rpath` could look something like:

```ruby
def delete_rpath(path, options = {})
  each_machos_with_check(options) do |macho|
    macho.delete_rpath(path)
  end
  synchronize_raw_data
end
```

Of course all those newly invented names are up for debate (I didn't think too long about them and as we all know naming things is hard).

woodruffw:
> Here, AppropriateBaseClass could potentially be ArgumentError and DylibUnknownError and similar exception classes could become children of ArgumentError.

Since we already have a (flat) `MachOError` hierarchy going on, I could make exceptions like `DylibUnknownError` all subclasses of `MachOModificationError` or something similar. This would allow us to be sufficiently generic while still distinguishing between API validation and modification failure.

> each_machos_with_check(options)

I think something like this would be ideal (perhaps just `each_macho`, given that the optional `options` parameter will be documented?)

UniqMartin:
Sounds all good to me! Let's get this idea implemented and then we can discuss the finer details with actual code in front of us. (I doubt we'll have to overthrow the entire design again.)

woodruffw:
> I could make exceptions like DylibUnknownError all subclasses of MachOModificationError or something similar.

After thinking about this, I think that it makes more sense to just rescue the toplevel `MachOError` exception - that way we'll catch mistakes that "succeed" are only exposed when the header/load commands are re-parsed.

UniqMartin:
> After thinking about this, I think that it makes more sense to just rescue the toplevel `MachOError` exception - that way we'll catch mistakes that "succeed" are only exposed when the header/load commands are re-parsed.

(I have to admit that I failed to fully parse this comment, thus forgive me if my comment is misguided.) This sounds like a dangerous path to take. We definitely don't want to rescue and silence all exceptions if they only occur for a single Mach-O slice out of possibly many slices.

An example: One slice cannot be modified because there's insufficient space in the Mach-O header padding, while all other slices have sufficient padding and thus the update succeeds there. This should be causing an error even in non-strict mode.

woodruffw:
> (I have to admit that I failed to fully parse this comment, thus forgive me if my comment is misguided.) This sounds like a dangerous path to take. We definitely don't want to rescue and silence all exceptions if they only occur for a single Mach-O slice out of possibly many slices.

You understood it correctly! I wasn't thinking, we definitely **don't** want to silence those types of errors.

lib/macho/exceptions.rb
```diff
@@ -163,6 +163,19 @@ def initialize(path)
...
+  # Raised when a modification fails one or more slices within a fat Mach-O.
+  class MachOSliceModificationError < MachOError
+    # @param strict [Boolean] the strictness used during the failed modification
+    # @param index [Fixnum] the index of the slice within the fat Mach-O
+    # @param cause [MachO::MachOModificationException] the inner exception
+    def initialize(strict, index, cause)
+      strictness = strict ? "strict" : "nonstrict"
+      inner_error = "#{cause.class}: #{cause}"
+
+      super "#{inner_error} during #{strictness} modification at slice #{index}"
```
UniqMartin: “[…] modification **of** slice […]”?

lib/macho/exceptions.rb
```diff
@@ -163,6 +163,19 @@ def initialize(path)
     end
   end
 
+  # Raised when a modification fails one or more slices within a fat Mach-O.
+  class MachOSliceModificationError < MachOError
+    # @param strict [Boolean] the strictness used during the failed modification
+    # @param index [Fixnum] the index of the slice within the fat Mach-O
+    # @param cause [MachO::MachOModificationException] the inner exception
```
UniqMartin: I guess this is supposed to be `MachO::MachOSliceModificationError`, right?

lib/macho/fat_file.rb
```diff
@@ -133,12 +133,15 @@ def dylib_id
 
     # Changes the file's dylib ID to `new_id`. If the file is not a dylib, does nothing.
     # @example
-    #  file.dylib_id = 'libFoo.dylib'
+    #  file.change_dylib_id('libFoo.dylib')
     # @param new_id [String] the new dylib ID
+    # @param options [Hash]
+    # @option options [Boolean] :strict (true) whether or not to fail loudly
+    #  with an exception if the change cannot be performed
```
UniqMartin: I think the wording of this documentation string needs to be tweaked a bit to be more specific and to better match the changed behavior we agreed on. (And of course this needs to change in all the other places this is used, too.)

woodruffw:
Using an encapsulating exception like `MachOSliceModificationError` breaks the isomorphism between `MachOFile` and `FatFile`, in the sense that the two no longer throw the same exception:

```ruby
macho.dylib_id = "x" * 4096 # HeaderPadError
fat.dylib_id = "x" * 4096 # MachOSliceModificationError
```

That's not a dealbreaker for me (since both are children of `MachOModificationError`, but it might be a source of confusion. Unfortunately, I don't really have an alternative that preserves the valuable information we get from being in the context of `FatFile` (i.e., `index`).

--

Besides that (very minor) nit, what do you think of renaming `MachOSliceModificationError` to `FatSliceModificationError`? I think that the latter more accurately reflects the semantics of the exception (a failure to modify a slice of a fat file, not a single-arch Mach-O).

UniqMartin:
> Using an encapsulating exception like `MachOSliceModificationError` breaks the isomorphism between `MachOFile` and `FatFile`, in the sense that the two no longer throw the same exception:
>
> ```ruby
> macho.dylib_id = "x" * 4096 # HeaderPadError
> fat.dylib_id = "x" * 4096 # MachOSliceModificationError
> ```

I can relate to this concern. Though you managed to pick an example that isn't related to the problem. :wink: I thought we had agreed that a `HeaderPadError` should always be a fatal condition, no matter whether strict or not and no matter if fat or thin binary. `MachOSliceModificationError` should only be wrapping exceptions that can be non-fatal under certain circumstances.

> That's not a dealbreaker for me (since both are children of `MachOModificationError`), but it might be a source of confusion. Unfortunately, I don't really have an alternative that preserves the valuable information we get from being in the context of `FatFile` (i.e., `index`).

We could also equip `MachOModificationError` with a `macho_slice` attribute that defaults to `nil` and gets set in `FatFile#each_macho` after rescuing the exception and before re-raising it. `MachOModificationError` could contain the necessary logic to adjust the message depending on whether `macho_slice` is set. (And I guess we'll need a `strictness` attribute, too.)

Now that this popped into my head, I think this is a superior approach to the error handling (or at least to how the exceptions are organized and re-raised). What do you think?

----

> Besides that (very minor) nit, what do you think of renaming `MachOSliceModificationError` to `FatSliceModificationError`? I think that the latter more accurately reflects the semantics of the exception (a failure to modify a slice of a fat file, not a single-arch Mach-O).

I guess this question is somewhat irrelevant, if you agree on the above. But my thinking was that `MachOSlice` (because it is only a slice of possibly many slices) reflects the information that it is part of a fat file. In contrast, `FatSlice` sounds to me like there is a fat slice somewhere, but there's actually none and the slice itself is thin (and part of a single fat file).

lib/macho/exceptions.rb
```diff
@@ -3,6 +3,10 @@ module MachO
   class MachOError < RuntimeError
   end
 
+  # A generic Mach-O error during file modification.
+  class MachOModificationError < MachOError
```
UniqMartin: Let's be super-explicit here. I think this should be something like “A generic Mach-O error during file modification, that is non-fatal and might be suppressed if operating in non-strict mode.” (I'm not particularly happy with this wording, but I hope the message gets across. Please feel free to phrase this better!)

We could also consider a name different from `MachOModificationError` that better reflects the nature of these exceptions.

lib/macho/exceptions.rb
```diff
@@ -115,15 +119,15 @@ def initialize(lc)
...
+  class OffsetInsertionError < MachOModificationError
     # @param offset [Fixnum] the invalid offset
     def initialize(offset)
       super "Insertion at offset #{offset} is not valid"
     end
   end
 
   # Raised when load commands are too large to fit in the current file.
-  class HeaderPadError < MachOError
+  class HeaderPadError < MachOModificationError
```
UniqMartin: As pointed out in a main comment, I don't think this should be reclassified. Please carefully review all the exceptions that can be raised by the modification methods and think whether they should be fatal or non-fatal (in non-strict mode).

lib/macho/fat_file.rb
```diff
@@ -312,6 +329,33 @@ def get_machos
...
+
+      machos.each_with_index do |macho, index|
+        begin
+          yield macho
+        rescue MachOModificationError => e
+          errors << [index, e]
+        end
+      end
+
+      strict = options.fetch(:strict, false)
```
UniqMartin: This disagrees with the documentation regarding the default for `:strict`. However, I think the default suggested by the documentation is ultimately the better choice (i.e. we should require an opt-out of the strict behavior) and for this reason, this should become `options.fetch(:strict, true)`.

lib/macho/macho_file.rb
```diff
@@ -305,23 +305,32 @@ def dylib_id
...
       raise ArgumentError.new("new ID must be a String") unless new_id.is_a?(String)
       return unless dylib?
 
       old_lc = command(:LC_ID_DYLIB).first
-      raise DylibIdMissingError.new unless old_lc
+
+      if old_lc.nil?
+        return unless options.fetch(:strict, true)
+        raise DylibIdMissingError.new
+      end
```
UniqMartin: This change needs to be reverted. The modification methods in `MachOFile` need to always raise an exception, independently of `options` (because `options` currently only matters if multiple Mach-O slices are involved).

lib/macho/macho_file.rb
```diff
@@ -337,11 +346,18 @@ def linked_dylibs
...
     # @raise [MachO::DylibUnknownError] if no shared library has the old name
-    def change_install_name(old_name, new_name)
+    def change_install_name(old_name, new_name, options = {})
       old_lc = dylib_load_commands.find { |d| d.name.to_s == old_name }
-      raise DylibUnknownError.new(old_name) if old_lc.nil?
+
+      if old_lc.nil?
+        return unless options.fetch(:strict, true)
+        raise DylibUnknownError.new(old_name)
+      end
```
UniqMartin: This change needs to be reverted. The modification methods in `MachOFile` need to always raise an exception, independently of `options` (because `options` currently only matters if multiple Mach-O slices are involved).

lib/macho/macho_file.rb
```diff
@@ -362,13 +378,25 @@ def rpaths
...
+
+      if rpaths.include?(new_path)
+        return unless options.fetch(:strict, true)
+        raise RpathExistsError.new(new_path)
+      end
+
+      if old_lc.nil?
+        return unless options.fetch(:strict, true)
+        raise RpathUnknownError.new(old_path)
+      end
```
UniqMartin: This change needs to be reverted. The modification methods in `MachOFile` need to always raise an exception, independently of `options` (because `options` currently only matters if multiple Mach-O slices are involved).

lib/macho/macho_file.rb
```diff
@@ -382,10 +410,16 @@ def change_rpath(old_path, new_path)
...
+    #  with an exception if the change cannot be performed
     # @return [void]
     # @raise [MachO::RpathExistsError] if the runtime path already exists
-    def add_rpath(path)
-      raise RpathExistsError.new(path) if rpaths.include?(path)
+    def add_rpath(path, options = {})
+      if rpaths.include?(path)
+        return unless options.fetch(:strict, true)
+        raise RpathExistsError.new(path)
+      end
```
UniqMartin: This change needs to be reverted. The modification methods in `MachOFile` need to always raise an exception, independently of `options` (because `options` currently only matters if multiple Mach-O slices are involved).

(The added `options` argument needs to stay of course.)

lib/macho/macho_file.rb
```diff
@@ -397,11 +431,18 @@ def add_rpath(path)
...
     # @raise [MachO::RpathUnknownError] if no such runtime path exists
-    def delete_rpath(path)
+    def delete_rpath(path, options = {})
       rpath_cmds = command(:LC_RPATH).select { |r| r.path.to_s == path }
-      raise RpathUnknownError.new(path) if rpath_cmds.empty?
+
+      if rpath_cmds.empty?
+        return unless options.fetch(:strict, true)
+        raise RpathUnknownError.new(path)
+      end
```
UniqMartin: This change needs to be reverted. The modification methods in `MachOFile` need to always raise an exception, independently of `options` (because `options` currently only matters if multiple Mach-O slices are involved).

test/test_fat.rb
```diff
@@ -269,7 +269,7 @@ def test_change_dylib_id
       really_big_id = "x" * 4096
 
       # test failsafe for excessively large IDs (w/ no special linking)
-      assert_raises MachO::HeaderPadError do
+      assert_raises MachO::MachOSliceModificationError do
```
UniqMartin: This needs to be reverted as this test should always result in a fatal `HeaderPadError` exception.

lib/macho/exceptions.rb
woodruffw: > We could also consider a name different from MachOModificationError that better reflects the nature of these exceptions.

This might get us uncomfortably close to checked and unchecked exceptions a la Java, but how about `MachONonFatalModificationError` (and `MachOFatalModificationError` for things like `HeaderPadError`)?

Agreed on the documentation :+1:

lib/macho/macho_file.rb
woodruffw: Sounds good, for this and all below. I'll also remove the `@option` documentation line now that we don't consider `strict` here.

lib/macho/exceptions.rb
woodruffw: I'm a little confused about how we're using "fatal" here, actually: do we mean "fatal" as in "halts the program" or as in "would do something very bad [to the Mach-O] if it were allowed to continue"?

woodruffw:
> We could also equip MachOModificationError with a macho_slice attribute that defaults to nil and gets set in FatFile#each_macho after rescuing the exception and before re-raising it. MachOModificationError could contain the necessary logic to adjust the message depending on whether macho_slice is set. (And I guess we'll need a strictness attribute, too.)

Could you describe the structure you have in mind? Sorry, I'm having a little trouble visualizing it :sweat_smile:

lib/macho/exceptions.rb
UniqMartin: Sorry for the confusion! I'm not entirely happy with the “fatal” term either but couldn't think of a better term. In this context I was using it to distinguish between modification errors that should always be raised (header pad error being one example) and modification errors that could be absorbed by our front-end API (e.g. `FatFile#delete_rpath`) in non-strict mode (e.g. RPATH deletion succeeds on one Mach-O slice but fails on another).

lib/macho/exceptions.rb
```diff
@@ -3,6 +3,14 @@ module MachO
   class MachOError < RuntimeError
   end
 
+  # A generic and non-fatal Mach-O error during file modification.
```
UniqMartin: I think we can drop `generic` from the description. It doesn't seem to add any value.

lib/macho/exceptions.rb
```diff
@@ -3,6 +3,14 @@ module MachO
...
+  class NonFatalModificationError < MachOError
+  end
+
+  # A generic and fatal Mach-O error during file modification.
```
UniqMartin: I think we can drop `generic` from the description. It doesn't seem to add any value.

lib/macho/macho_file.rb
```diff
@@ -362,14 +366,26 @@ def rpaths
     #  file.change_rpath("/usr/lib", "/usr/local/lib")
     # @param old_path [String] the old runtime path
     # @param new_path [String] the new runtime path
+    # @param options [Hash]
+    # @return [void]
```
UniqMartin: Something went wrong here. (Note that there's already a `@return [void]` directly below this line.)

lib/macho/macho_file.rb
UniqMartin: You brought back the original checks further up in this method but still left this new code intact?

woodruffw: Oops...

lib/macho/exceptions.rb
woodruffw: No worries! I was also confusing myself my reusing the term in the exceptions. How do you feel about `CheckedModificationError` and `UncheckedModificationError` instead? That gets us away from the connotation of program halting and more accurately conveys the fact that we're checking and filtering subclasses of the particular error.

UniqMartin:
Pulling this out from https://github.com/Homebrew/ruby-macho/pull/55#discussion_r73792829 so it's not completely hidden (and lost):

> No worries! I was also confusing myself my reusing the term in the exceptions. How do you feel about `CheckedModificationError` and `UncheckedModificationError` instead? That gets us away from the connotation of program halting and more accurately conveys the fact that we're checking and filtering subclasses of the particular error.

That's a nice suggestion; I think that's terminology we could borrow. I went looking for something similar in the Ruby world, but I didn't find anything.

Another idea I just came up with is to use `ModificationError` and `ModificationWarning`. The only problem I have with this is that having something named `*Warning` in a hierarchy of exception classes feels a bit weird. What we're effectively doing with the strict behavior for fat binaries is to promote warnings to errors. (Maybe `RecoverableModificationError` is better?) In any case, and just to be sure we're on the same page, here's the exception hierarchy I'm envisioning:

```
-- MachOError
   |-- (various other exceptions)
   \-- ModificationError
       |-- RecoverableModificationError (ModificationWarning)
       |   |-- DylibUnknownError
       |   |-- DylibIdMissingError
       |   |-- RpathUnknownError
       |   \-- RpathExistsError
       |-- OffsetInsertionError
       \-- HeaderPadError
```

Does this look reasonable?

> Could you describe the structure you have in mind? Sorry, I'm having a little trouble visualizing it 😅

Will do. Give me a few moments to prepare some demo code.

UniqMartin:
> Could you describe the structure you have in mind? Sorry, I'm having a little trouble visualizing it 😅

Code (changes on top of this branch): https://github.com/UniqMartin/ruby-macho/commit/cfa5e91b0a5ca5f48693781decdaba74f03053f9 in my [`strictness` branch](https://github.com/UniqMartin/ruby-macho/tree/strictness)

Preparations:

```console
$ cp test/bin/i386/hello.bin hello-i386
$ cp test/bin/x86_64/hello.bin hello-x86_64
$ lipo -create -output hello-consistent hello-i386 hello-x86_64
$ install_name_tool -add_rpath /mesa/special hello-x86_64
$ lipo -create -output hello-inconsistent hello-i386 hello-x86_64
$ ruby -Ilib -rmacho -e 'puts MachO.open(ARGV.shift).machos.map(&:rpaths).inspect' hello-consistent
[["made_up_path"], ["made_up_path"]]
$ ruby -Ilib -rmacho -e 'puts MachO.open(ARGV.shift).machos.map(&:rpaths).inspect' hello-inconsistent
[["made_up_path"], ["made_up_path", "/mesa/special"]]
```

Tests (the important difference and the point of this PR is the treatment of `hello-inconsistent`):

```console
$ # Strict mode:
$ add-rpath() { ruby -Ilib -rmacho -e 'MachO.open(ARGV.shift).add_rpath(*ARGV, :strict => true) rescue puts $!.inspect' "$@"; }
$ add-rpath hello-inconsistent /mesa/special
#<MachO::RpathExistsError: While modifying Mach-O slice 1: /mesa/special already exists>

$ # … no change for consistent fat binaries:
$ add-rpath hello-consistent made_up_path
#<MachO::RpathExistsError: While modifying Mach-O slice 0: made_up_path already exists>
$ add-rpath hello-consistent /mesa/special

$ # … no change for non-fat binaries:
$ add-rpath hello-i386 made_up_path
#<MachO::RpathExistsError: made_up_path already exists>
$ add-rpath hello-i386 /mesa/special

$ # Non-strict mode:
$ add-rpath() { ruby -Ilib -rmacho -e 'MachO.open(ARGV.shift).add_rpath(*ARGV, :strict => false) rescue puts $!.inspect' "$@"; }
$ add-rpath hello-inconsistent /mesa/special

$ # … no change for consistent fat binaries:
$ add-rpath hello-consistent made_up_path
#<MachO::RpathExistsError: While modifying Mach-O slice 0: made_up_path already exists>
$ add-rpath hello-consistent /mesa/special

$ # … no change for non-fat binaries:
$ add-rpath hello-i386 made_up_path
#<MachO::RpathExistsError: made_up_path already exists>
$ add-rpath hello-i386 /mesa/special
```

woodruffw:
> Another idea I just came up with is to use ModificationError and ModificationWarning. The only problem I have with this is that having something named *Warning in a hierarchy of exception classes feels a bit weird. What we're effectively doing with the strict behavior for fat binaries is to promote warnings to errors. (Maybe RecoverableModificationError is better?)

Yeah, I think `ModificationWarning` will ultimately become confusing (since single-arch Mach-Os will be raising subclasses of it and *not* filtering them before getting to the user). `RecoverableModificationError` gets us to the same meaning as `CheckedModificationError`, with the added benefit of not being mutually exclusive (like `CheckedModificationError` is with `UncheckedModificationError`) :+1: 

> In any case, and just to be sure we're on the same page, here's the exception hierarchy I'm envisioning

Yep!

woodruffw:
> Code (changes on top of this branch): UniqMartin@cfa5e91 in my strictness branch

This looks great! Re-raising the exception immediately with the added context feels much more natural to me. 

I cherry-picked your commit right on top of this branch - do you know how to preserve authorship (if possible) during a squash?

UniqMartin:
> I cherry-picked your commit right on top of this branch - do you know how to preserve authorship (if possible) during a squash?

I guess there must be a more elegant solution, but this works:

```console
$ git rebase -i HEAD~2 # squash as usual
$ git commit --no-edit --amend --author="$(git show -s --format='%an <%ae>' origin/lc-modification-strictness)" # steal authorship information from commit authored by me
```

UniqMartin:
I feel a bit bad stealing your commit credit as you've done the bulk of the work, though. 😳

woodruffw:
> I feel a bit bad stealing your commit credit as you've done the bulk of the work, though.

Don't worry about it; although we could probably do something like `--author="William Woodruff <william@tuffbizz.com> and Martin Afanasjew <martin@afanasjew.de>"`. I'm not sure if GitHub would like that, though.

Either way, I still show up as the PR's author and committer, so I'm fine with it :smile:

UniqMartin:
> Either way, I still show up as the PR's author and committer, so I'm fine with it 😄

Then please go ahead and squash-merge this yourself, assuming you're happy with the PR. (I think it's ready.) Otherwise if I do this myself, you'll be robbed of your committer credit, too.

woodruffw:
Done! :tada:

UniqMartin:
Awesome! :tada: Thanks for surviving through my reviews and constant complaints. :joy:

PR #56, opened 2016-08-01
General refactoring.

woodruffw:
Another low-priority PR, just cleans up the code a bit and adds a `.rubocop.yml` for future style enforcement.

`rubocop` complains about roughly 100 other things in the `lib/` tree, although most of them are even less significant (methods over 10 lines, for example).
All of the unresolved violations are here: [ruby-macho-rubocop.txt](https://github.com/Homebrew/ruby-macho/files/392754/ruby-macho-rubocop.txt)

cc @UniqMartin

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,53 @@
+Style/StringLiterals:
+  EnforcedStyle: double_quotes
+
+Style/StringLiteralsInInterpolation:
+  EnforcedStyle: double_quotes
+
+Style/TrailingCommaInLiteral:
+  EnforcedStyleForMultiline: comma
+
+Style/FormatString:
+  EnforcedStyle: percent
+
+# we prefer compact if-else-end/case-when-end alignment
+Lint/EndAlignment:
+  AlignWith: variable
+
+Style/CaseIndentation:
+  IndentWhenRelativeTo: end
+
+# Ruby 1.8 compatibility
+Style/DotPosition:
+  EnforcedStyle: trailing
+
+Style/HashSyntax:
+  EnforcedStyle: hash_rockets
+
+Style/TrailingCommaInArguments:
+  EnforcedStyleForMultiline: no_comma
+
+# TODO: re-enable in the future
+Metrics/LineLength:
+  Enabled: false
+
+Metrics/AbcSize:
+  Enabled: false
+
+Metrics/ClassLength:
+  Enabled: false
+
+Metrics/CyclomaticComplexity:
+  Enabled: false
+
+Metrics/MethodLength:
+  Enabled: false
+
+Metrics/ModuleLength:
+  Enabled: false
+
+Metrics/ParameterLists:
+  Enabled: false
+
+Metrics/PerceivedComplexity:
+  Enabled: false
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -146,9 +146,9 @@ module MachO
   class HeaderPadError < ModificationError
     # @param filename [String] the filename
     def initialize(filename)
-      super "Updated load commands do not fit in the header of " +
-      "#{filename}. #{filename} needs to be relinked, possibly with " +
-      "-headerpad or -headerpad_max_install_names"
+      super "Updated load commands do not fit in the header of " \
+        "#{filename}. #{filename} needs to be relinked, possibly with " \
+        "-headerpad or -headerpad_max_install_names"
     end
   end
 
--- a/lib/macho/headers.rb
+++ b/lib/macho/headers.rb
@@ -32,8 +32,8 @@ module MachO
     MH_MAGIC => "MH_MAGIC",
     MH_CIGAM => "MH_CIGAM",
     MH_MAGIC_64 => "MH_MAGIC_64",
-    MH_CIGAM_64 => "MH_CIGAM_64"
-  }
+    MH_CIGAM_64 => "MH_CIGAM_64",
+  }.freeze
 
   # mask for CPUs with 64-bit architectures (when running a 64-bit ABI?)
   # @api private
@@ -85,7 +85,7 @@ module MachO
     CPU_TYPE_ARM64 => :arm64,
     CPU_TYPE_POWERPC => :ppc,
     CPU_TYPE_POWERPC64 => :ppc64,
-  }
+  }.freeze
 
   # mask for CPU subtype capabilities
   # @api private
@@ -114,7 +114,7 @@ module MachO
 
   # @see CPU_SUBTYPE_586
   # @api private
-  CPU_SUBTYPE_PENT  = CPU_SUBTYPE_586
+  CPU_SUBTYPE_PENT = CPU_SUBTYPE_586
 
   # the Pentium Pro (P6) sub-type for `CPU_TYPE_I386`
   # @api private
@@ -412,7 +412,7 @@ module MachO
     MH_DYLIB_STUB => :dylib_stub,
     MH_DSYM => :dsym,
     MH_KEXT_BUNDLE => :kext_bundle,
-  }
+  }.freeze
 
   # association of mach header flag symbols to values
   # @api private
@@ -442,8 +442,8 @@ module MachO
     :MH_DEAD_STRIPPABLE_DYLIB => 0x400000,
     :MH_HAS_TLV_DESCRIPTORS => 0x800000,
     :MH_NO_HEAP_EXECUTION => 0x1000000,
-    :MH_APP_EXTENSION_SAFE => 0x02000000
-  }
+    :MH_APP_EXTENSION_SAFE => 0x02000000,
+  }.freeze
 
   # Fat binary header structure
   # @see MachO::FatArch
@@ -457,7 +457,7 @@ module MachO
     # always big-endian
     # @see MachOStructure::FORMAT
     # @api private
-    FORMAT = "N2"
+    FORMAT = "N2".freeze
 
     # @see MachOStructure::SIZEOF
     # @api private
@@ -492,7 +492,7 @@ module MachO
     # always big-endian
     # @see MachOStructure::FORMAT
     # @api private
-    FORMAT = "N5"
+    FORMAT = "N5".freeze
 
     # @see MachOStructure::SIZEOF
     # @api private
@@ -533,7 +533,7 @@ module MachO
 
     # @see MachOStructure::FORMAT
     # @api private
-    FORMAT = "L=7"
+    FORMAT = "L=7".freeze
 
     # @see MachOStructure::SIZEOF
     # @api private
@@ -541,7 +541,7 @@ module MachO
 
     # @api private
     def initialize(magic, cputype, cpusubtype, filetype, ncmds, sizeofcmds,
-        flags)
+                   flags)
       @magic = magic
       @cputype = cputype
       # For now we're not interested in additional capability bits also to be
@@ -571,7 +571,7 @@ module MachO
 
     # @see MachOStructure::FORMAT
     # @api private
-    FORMAT = "L=8"
+    FORMAT = "L=8".freeze
 
     # @see MachOStructure::SIZEOF
     # @api private
@@ -579,7 +579,7 @@ module MachO
 
     # @api private
     def initialize(magic, cputype, cpusubtype, filetype, ncmds, sizeofcmds,
-        flags, reserved)
+                   flags, reserved)
       super(magic, cputype, cpusubtype, filetype, ncmds, sizeofcmds, flags)
       @reserved = reserved
     end
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -172,7 +172,7 @@ module MachO
 
     # @see MachOStructure::FORMAT
     # @api private
-    FORMAT = "L=2"
+    FORMAT = "L=2".freeze
 
     # @see MachOStructure::SIZEOF
     # @api private
@@ -186,16 +186,16 @@ module MachO
       bin = view.raw_data.slice(view.offset, bytesize)
       format = Utils.specialize_format(self::FORMAT, view.endianness)
 
-      self.new(view, *bin.unpack(format))
+      new(view, *bin.unpack(format))
     end
 
     # Creates a new (viewless) command corresponding to the symbol provided
     # @param cmd_sym [Symbol] the symbol of the load command being created
     # @param args [Array] the arguments for the load command being created
     def self.create(cmd_sym, *args)
-      raise LoadCommandNotCreatableError.new(cmd_sym) unless CREATABLE_LOAD_COMMANDS.include?(cmd_sym)
+      raise LoadCommandNotCreatableError, cmd_sym unless CREATABLE_LOAD_COMMANDS.include?(cmd_sym)
 
-      klass = MachO.const_get "#{LC_STRUCTURES[cmd_sym]}"
+      klass = MachO.const_get LC_STRUCTURES[cmd_sym]
       cmd = LOAD_COMMAND_CONSTANTS[cmd_sym]
 
       # cmd will be filled in, view and cmdsize will be left unpopulated
@@ -227,7 +227,7 @@ module MachO
     #  if the load command can't be serialized
     # @api private
     def serialize(context)
-      raise LoadCommandNotSerializableError.new(LOAD_COMMANDS[cmd]) unless serializable?
+      raise LoadCommandNotSerializableError, LOAD_COMMANDS[cmd] unless serializable?
       format = Utils.specialize_format(FORMAT, context.endianness)
       [cmd, SIZEOF].pack(format)
     end
@@ -243,7 +243,7 @@ module MachO
       LOAD_COMMANDS[cmd]
     end
 
-    alias :to_sym :type
+    alias to_sym type
 
     # @return [String] a string representation of the load command's identifying number
     def to_s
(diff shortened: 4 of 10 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,6 @@
+StringLiterals:
+  Enabled: false
```
UniqMartin: You're missing the `Style/` prefix in the preceding line and instead of simply disabling the check, it's better to reconfigure it to our preferred style:

```yaml
Style/StringLiterals:
  EnforcedStyle: double_quotes
```

You also most likely want to add the following for consistency with the above and with Homebrew code in general:

```yaml
Style/StringLiteralsInInterpolation:
  EnforcedStyle: double_quotes
```

.rubocop.yml
```diff
@@ -0,0 +1,6 @@
...
+
+LineLength:
```
UniqMartin: This should be `Metrics/LineLength:`.

lib/macho/headers.rb
```diff
@@ -33,7 +33,7 @@ module MachO
     MH_CIGAM => "MH_CIGAM",
     MH_MAGIC_64 => "MH_MAGIC_64",
     MH_CIGAM_64 => "MH_CIGAM_64"
```
UniqMartin: If this missing trailing comma wasn't pointed out, then I guess we should add the following:

```yaml
Style/TrailingCommaInLiteral:
  EnforcedStyleForMultiline: comma
```

lib/macho/headers.rb
```diff
@@ -412,38 +412,38 @@ module MachO
...
+    MH_BINDS_TO_WEAK: 0x10000,
+    MH_ALLOW_STACK_EXECUTION: 0x20000,
+    MH_ROOT_SAFE: 0x40000,
+    MH_SETUID_SAFE: 0x80000,
+    MH_NO_REEXPORTED_DYLIBS: 0x100000,
+    MH_PIE: 0x200000,
+    MH_DEAD_STRIPPABLE_DYLIB: 0x400000,
+    MH_HAS_TLV_DESCRIPTORS: 0x800000,
+    MH_NO_HEAP_EXECUTION: 0x1000000,
+    MH_APP_EXTENSION_SAFE: 0x02000000
```
UniqMartin: I'm not happy to give up Ruby 1.8 compatibility just yet and this change makes restoring compatibility a lot harder. To that end, would you mind adding the following to the `.rubocop.yml` and reverting this and similar changes?

```yaml
# Ruby 1.8 compatibility
Style/DotPosition:
  EnforcedStyle: trailing
Style/HashSyntax:
  EnforcedStyle: hash_rockets
Style/TrailingCommaInArguments:
  EnforcedStyleForMultiline: no_comma
```

lib/macho/macho_file.rb
```diff
@@ -556,7 +556,7 @@ def low_fileoff
 
       segments.each do |seg|
         sections(seg).each do |sect|
-          next if sect.size == 0
+          next if sect.size.zero?
```
UniqMartin: Was this actually pointed out by RuboCop? It feels like this change reduces readability without any clear benefit. And the actual fix should most likely be `sect.empty?`.

.rubocop.yml
UniqMartin: Also, I'm fine with leaving this switched off for the time being, but I think we should aim at re-enabling this check at a later time (in the not too distant future) and make sure we stay within an 80-character boundary.

woodruffw: Definitely in favor of that!

lib/macho/exceptions.rb
```diff
@@ -127,8 +127,8 @@ class HeaderPadError < MachOError
     # @param filename [String] the filename
     def initialize(filename)
       super "Updated load commands do not fit in the header of " +
-      "#{filename}. #{filename} needs to be relinked, possibly with " +
-      "-headerpad or -headerpad_max_install_names"
+        "#{filename}. #{filename} needs to be relinked, possibly with " +
+        "-headerpad or -headerpad_max_install_names"
```
UniqMartin: RuboCop gives good advice here. Aside from the indentation fix, here's what you want to do:

```ruby
super "Updated load commands do not fit in the header of " \
  "#{filename}. #{filename} needs to be relinked, possibly with " \
  "-headerpad or -headerpad_max_install_names"
```

Concatenation of the strings with `String#+` will happen every time this code is executed, but if you're using line continuations (`\`) it will be the same as `"hello " "world"` (like in C) and the concatenation will already happen at parse time.

lib/macho/macho_file.rb
woodruffw: RuboCop actually pointed out both as alternatives - I just went with the first, although `sect.empty?` makes a lot more sense :+1:

.rubocop.yml
UniqMartin: The second part will also take care of some of the complaints that remained in the RuboCop output.

woodruffw:
Sorry for the rookie mistakes regarding RuboCop - this is my first time configuring/using it :sweat_smile:.

lib/macho/headers.rb
woodruffw: Sure!

lib/macho/exceptions.rb
woodruffw: Sounds good!

I ignored that suggestion originally because I generally shy away from metasyntax like line continuations, but best practice is best practice :smile:

UniqMartin:
> Sorry for the rookie mistakes regarding RuboCop - this is my first time configuring/using it

No worries, I'm happy to point you into the right direction (or what I believe to be the right direction). I certainly had more experience tweaking those rules in the context of Homebrew. Overall I'm very happy that you came up with this PR on your own. :+1:

UniqMartin:
> `rubocop` complains about roughly 100 other things in the `lib/` tree, although most of them are even less significant (methods over 10 lines, for example). All of the unresolved violations are here: [ruby-macho-rubocop.txt](https://github.com/Homebrew/ruby-macho/files/392754/ruby-macho-rubocop.txt)

Some of them will already vanish after addressing my comments. Others, like the length and complexity metrics are something I'm less convinced of (or at least I find the limits to be quite harsh), thus I'd be personally alright with:

```yaml
Metrics/AbcSize:
  Enabled: false
Metrics/ClassLength:
  Enabled: false
Metrics/CyclomaticComplexity:
  Enabled: false
Metrics/MethodLength:
  Enabled: false
Metrics/ModuleLength:
  Enabled: false
Metrics/ParameterLists:
  Enabled: false
Metrics/PerceivedComplexity:
  Enabled: false
```

We can sort out most of the remaining items either in the next iteration of this PR or just ignore them for the time being and deal with them in a later PR.

Pro Tip: Use `rubocop -DES` for extra details like the cop name responsible for the output and links to the corresponding section of the style guide. I also like to append `-f progress -f offenses` to the argument list to get an additional summary of the various types of offenses that were raised.

woodruffw:
> Metrics/AbcSize:
>   Enabled: false
> Metrics/ClassLength:
>   Enabled: false
> Metrics/CyclomaticComplexity:
>   Enabled: false
> Metrics/MethodLength:
>   Enabled: false
> Metrics/ModuleLength:
>   Enabled: false
> Metrics/ParameterLists:
>   Enabled: false
> Metrics/PerceivedComplexity:
>   Enabled: false

This looks good to me! I'm going to make all the changes you've recommended tonight.

Thanks for the tip about the additional RuboCop flags :smile:

lib/macho/macho_file.rb
woodruffw: Actually, it looks like RuboCop mistakenly thought that `sect` was an array/enumerable here - it's actually just a `Section{,64}` object, and so it doesn't have an `#empty?` method. I can define one, if you prefer that over `size.zero?`.

woodruffw:
RuboCop still complains about unaligned parameters/arrays/cases, so I'm probably going to add those to the `.rubocop.yml` unless you think we should address them here.

There's also the matter of the unfortunately named `MachOFile#set_ncmds` and `MachOFile#set_sizeofcmds`. How do `update_ncmds` and `update_sizeofcmds` sound to you? Alternatively, we could do something like `raw_ncmds=`...

lib/macho/macho_file.rb
UniqMartin: You're right, RuboCop misinterprets this because it cannot know what the contents of `sect` will be, but it assumes that `.size` is indicative of a container. I failed to realize this when commenting on this. I see several viable options:

- Put a `# rubocop:disable Style/ZeroLengthPredicate` comment at the end of the line to locally silence the misguided advice.
- Add an `empty?` convenience method to `Section[64]`. (It makes sense to me to speak of an empty section if the section has a size of zero bytes.)
- Slightly diverge from the C struct and rename the `size` attribute.

UniqMartin:
> RuboCop still complains about unaligned parameters/arrays/cases, so I'm probably going to add those to the `.rubocop.yml` unless you think we should address them here.

I just went ahead and pushed a few commits that address these, partially to see for myself how this will change the code. I'd say that we should address them (I only focused my attention on the `lib/` directory; it is clean except for one documentation complaint). Feel free to rearrange, drop, or squash any of my commits that I so boldly pushed to your branch.

> There's also the matter of the unfortunately named `MachOFile#set_ncmds` and `MachOFile#set_sizeofcmds`. How do `update_ncmds` and `update_sizeofcmds` sound to you?

I'm in favor of `update_*`. That's already the verb we use in the documentation comment.

lib/macho/macho_file.rb
```diff
@@ -420,13 +420,13 @@ def delete_rpath(path)
...
-      else
-        raise ArgumentError, "not a valid segment"
-      end
+                      when MachO::SegmentCommand
+                        MachO::Section
+                      when MachO::SegmentCommand64
+                        MachO::Section64
+                      else
+                        raise ArgumentError, "not a valid segment"
+                      end
```
UniqMartin: If this looks unappealing to you, we could steal the following snippet from Homebrew's `.rubocop.yml` to gain a bit more flexibility with these:

```yaml
# we prefer compact if-else-end/case-when-end alignment
Lint/EndAlignment:
  AlignWith: variable
Style/CaseIndentation:
  IndentWhenRelativeTo: end
```

woodruffw:
Looks great to me! I'll squash these down into the RuboCop and general refactoring commits in a moment.

> I only focused my attention on the lib/ directory; it is clean except for one documentation complaint)

As did I. It might make sense at some point in the future to align the unit tests with our RuboCop enforced style, but that's not particularly high on my list of priorities for `ruby-macho` :smile:

lib/macho/macho_file.rb
woodruffw: Yeah, I do generally prefer the original form.
I'll add those instructions to the `.rubocop.yml`.

lib/macho/macho_file.rb
```diff
@@ -556,7 +560,7 @@ def low_fileoff
 
       segments.each do |seg|
         sections(seg).each do |sect|
-          next if sect.size == 0
+          next if sect.size == 0 # rubocop:disable Style/ZeroLengthPredicate
```
woodruffw: Looks like RuboCop still complains here about a `Style/NumericPredicate` violation. I'd personally just prefer `zero?` here, since it's equally readable (IMO, of course) and we don't have to disable two checks in-line.

lib/macho/macho_file.rb
woodruffw: > Add an empty? convenience method to Section[64]. (It makes sense to me to speak of an empty section if the section has a size of zero bytes.)

This sounds good to me!

UniqMartin: :+1:

> - Put a `# rubocop:disable Style/ZeroLengthPredicate` comment at the end of the line to locally silence the misguided advice.

I was still on RuboCop 0.41.2. That's why I didn't notice this was now causing multiple complaints and this suggestion isn't exactly practical.

lib/macho/macho_file.rb
UniqMartin: :+1:

I can see arguments for both styles, but I'm happy with either (but more used to the unindented style due to my Homebrew work).

lib/macho/macho_file.rb
UniqMartin: :+1: *To quote from the other related code comment:*

> I was still on RuboCop 0.41.2. That's why I didn't notice this was now causing multiple complaints and this suggestion isn't exactly practical.

UniqMartin:
> > I only focused my attention on the lib/ directory; it is clean except for one documentation complaint)
>
> As did I. It might make sense at some point in the future to align the unit tests with our RuboCop enforced style, but that's not particularly high on my list of priorities for `ruby-macho` :smile: 

Great! I think it's sufficient to tweak those style issues incrementally as the code gets modified (or maybe in a follow-up PR for the test suite as a whole). A good way to improve incrementally is to let your editor highlight RuboCop style violations for you while coding. I'm using [SublimeLinter](http://www.sublimelinter.com/) with its RuboCop plugin for that and loving it.

.rubocop.yml
```diff
@@ -0,0 +1,53 @@
...
+Style/DotPosition:
+  EnforcedStyle: trailing
+
+Style/HashSyntax:
+  EnforcedStyle: hash_rockets
+
+Style/TrailingCommaInArguments:
+  EnforcedStyleForMultiline: no_comma
+
+# TODO: disable in the future
```
UniqMartin: Should that be “re-enable in the future”?

woodruffw:
> A good way to improve incrementally is to let your editor highlight RuboCop style violations for you while coding. I'm using SublimeLinter with its RuboCop plugin for that and loving it.

Thanks! I knew about SublimeLinter, but I didn't know about its RuboCop plugin. I'll install it now :smile:

.rubocop.yml
woodruffw: Yep. I was thinking disable as in "disable these lines", but re-enable makes much more sense.

lib/macho/load_commands.rb
```diff
@@ -270,8 +270,8 @@ def initialize(lc, lc_str)
           lc_str_abs = view.offset + lc_str
           lc_end = view.offset + lc.cmdsize - 1
           raw_string = view.raw_data.slice(lc_str_abs..lc_end)
-          @string, null_byte, _ = raw_string.partition("\x00")
-          raise LCStrMalformedError.new(lc) if null_byte.empty?
+          @string, null_byte, = raw_string.partition("\x00")
```
UniqMartin: I'm not liking this resolution. The trailing comma on the left-hand side is too easy to overlook. How about the following?

```ruby
@string, null_byte, _padding = raw_string.partition("\x00")
```

woodruffw: I was thinking the same thing. `_padding` sounds good to me!

woodruffw:
Rebased; I'm happy with the state of this PR if you are :tada:

UniqMartin:
Rebased … and checked whether `rubocop` complains about any of the new/changed code? :wink:

woodruffw:
Hmm, RuboCop doesn't like that we never use the `options` arguments to the modification methods in`MachOFile`. I guess I'll make them `_options` if (until?) we end up needing them.

UniqMartin:
Could you try and adopt a workflow that looks something like this?

1. Make a change.
2. Self-review and basic sanity check (in this case re-running `rubocop`).
3. Push change.

Sorry if I'm wronging you with this, but it feels like you're sometimes skipping the second step. I don't think I should be pointing out stuff like in the following terminal output and we could be both more productive if I didn't have to, not only because of the incurred roundtrip delay.

```console
$ rubocop -D -f simple lib
== lib/macho/load_commands.rb ==
C:403:  1: Style/EmptyLines: Extra blank line detected.
== lib/macho/open.rb ==
C:  1:  1: Style/Documentation: Missing top-level module documentation comment.

12 files inspected, 2 offenses detected
```

(And if in doubt: No, I'm not referring to the second offense that has been there from the start. I'm talking about the other one that has been introduced due to the rebase.)

woodruffw:
> Sorry if I'm wronging you with this

You're not, I haven't been proactive about checking work that I *think* is correct. Sorry for imposing the extra burden on you, it's not fair of me to do.

(That last offense looks like another one of RuboCop's screw-ups - the `MachO` module is documented in `macho.rb`.)

UniqMartin:
Thanks! :bow:

> (That last offense looks like another one of RuboCop's screw-ups - the `MachO` module is documented in `macho.rb`.)

We can look into this later. I'm not 100% sure how RuboCop is performing the check, but I'm relatively confident it works on a file by file basis and thus cannot correlate information from different files. (It's perfectly normal to let it check one file only.) Since modules and classes can be reopened any time, I suspect it's using some heuristics to determine whether the module is new (and thus needs to be documented) or is just being reopened and added to …

woodruffw:
Yeah, that would make sense. For the time being I'm happy with just ignoring that line in `rubocop`'s output (disabling it probably wouldn't be good, since we don't want to leave future modules/classes accidentally undocumented).

UniqMartin:
Yes, agreed. It's not exactly bothering me and we can ignore it for the time being.

Slightly off topic, but I guess you'd like to cut a new release soon? Even if it's not 1.0 time yet, I think we have accumulated quite a few important changes since the last release and it would make sense to cut a new one. I think it would also makes sense to bring this new release to Homebrew proper, so that the test bot can throw some real world binaries at it and we can see how this goes, not to mention that we'll need to make some minor adjustments to the code that is bridging between the Homebrew-internal API and the `ruby-macho` API.

woodruffw:
I was just about to! It'll be up in a moment.

woodruffw:
I'll also open a PR on `brew` to bring the updated `ruby-macho` in and update our interfacing code.

PR #57, opened 2016-08-07
Deprecate MachOFile#sections in favor of SegmentCommand#sections.

woodruffw:
New code should always prefer `SegmentCommand#sections` over `MachOFile#sections`, as this better expresses the relationship between segments and their sections (and the parent Mach-O).

I'm leaving `MachOFile#sections` in for the time being (with a `@deprecated` warning), although all internal references to it have been removed. It might make sense to remove it before 1.0, though.

cc @UniqMartin

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -400,6 +400,24 @@ module MachO
       @flags = flags
     end
 
+
+    # All sections referenced within this segment.
+    # @return [Array<MachO::Section>] if the Mach-O is 32-bit
+    # @return [Array<MachO::Section64>] if the Mach-O is 64-bit
+    def sections
+      klass = case self
+      when MachO::SegmentCommand64
+        MachO::Section64
+      when MachO::SegmentCommand
+        MachO::Section
+      end
+
+      bins = view.raw_data[view.offset + self.class.bytesize, nsects * klass.bytesize]
+      bins.unpack("a#{klass.bytesize}" * nsects).map do |bin|
+        klass.new_from_bin(view.endianness, bin)
+      end
+    end
+
     # @example
     #  puts "this segment relocated in/to it" if sect.flag?(:SG_NORELOC)
     # @param flag [Symbol] a segment flag symbol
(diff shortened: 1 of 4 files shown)


lib/macho/load_commands.rb
```diff
@@ -400,6 +400,32 @@ def initialize(view, cmd, cmdsize, segname, vmaddr, vmsize, fileoff,
...
+    def sections
+      section_klass = case self
+      when MachO::SegmentCommand64
+        MachO::Section64
+      when MachO::SegmentCommand
+        MachO::Section
+      end
+
+      sections = []
+      return sections if nsects.zero?
```
UniqMartin: You can move this to the very top of the method and make it simply `return [] if nsects.zero?`.

lib/macho/load_commands.rb
```diff
@@ -400,6 +400,32 @@ def initialize(view, cmd, cmdsize, segname, vmaddr, vmsize, fileoff,
...
+
+      offset = view.offset + self.class.bytesize
+
+      nsects.times do
+        section_bin = view.raw_data[offset, section_klass.bytesize]
+        sections << section_klass.new_from_bin(view.endianness, section_bin)
+        offset += section_klass.bytesize
+      end
```
UniqMartin: You can completely eliminate the local variable `sections` by using `Enumerable#map` and putting this at the very end of the method (so the return value of the block is returned):

```ruby
nsects.times.map do
  section_bin = view.raw_data[offset, section_klass.bytesize]
  offset += section_klass.bytesize
  section_klass.new_from_bin(view.endianness, section_bin)
end
```

And an even more functional approach that completely avoids offset calculations would be:

```ruby
sections_bin = view.raw_data[offset, nsetcs * section_klass.bytesize]
sections_bin.each_slice(section_klass.bytesize).map do |section_bin|
  section_klass.new_from_bin(view.endianness, section_bin)
end
```

UniqMartin:
A few suggestions but overall nice cleanup! :+1:

lib/macho/load_commands.rb
woodruffw: I really like the second option, but I don't think there's a `String#each_slice` (or an efficient way to emulate one) :disappointed:

UniqMartin: How silly of me to not realize we're dealing with a `String` here! Sorry! Not sure how you feel about it, but my suggestion can be rescued with `String#unpack`:

```ruby
sections_bin = view.raw_data[offset, nsects * section_klass.bytesize]
sections_bin.unpack("a#{section_klass.bytesize}" * nsects).map do |section_bin|
  section_klass.new_from_bin(view.endianness, section_bin)
end
```

(I guess I'm equally fine with this version and what we currently have in the PR.)

woodruffw: I'll think I'll switch to this - getting rid of the `offset` variable is nice and that dynamic unpack is a nice touch :+1:

lib/macho/load_commands.rb
```diff
@@ -400,6 +400,24 @@ def initialize(view, cmd, cmdsize, segname, vmaddr, vmsize, fileoff,
...
+    # @return [Array<MachO::Section64>] if the Mach-O is 64-bit
+    def sections
+      section_klass = case self
+      when MachO::SegmentCommand64
+        MachO::Section64
+      when MachO::SegmentCommand
+        MachO::Section
+      end
+
+      sections_bin = view.raw_data[offset + self.class.bytesize, nsects * section_klass.bytesize]
```
UniqMartin: `offset` should be `view.offset` to avoid using the deprecated `LoadCommand#offset` method. And it would be nice to keep the length of the line manageable. One possible solution, since everything in here is about sections, would be to strip the prefix and use `klass`, `bins`, and `bin`.

woodruffw: > offset should be view.offset to avoid using the deprecated LoadCommand#offset method

Thanks for reminding me!



> One possible solution, since everything in here is about sections, would be to strip the prefix and use klass, bins, and bin.

:+1:

UniqMartin:
:tada:

woodruffw:
:clap:

PR #59, opened 2016-08-08
test: Replace "pass" with "skip".

woodruffw:
This makes the test output a little less pretty, but more accurately reflects the fact that we're skipping tests of unimplemented behavior.

cc @UniqMartin


UniqMartin:
Definitely the right thing to do. Thanks for rectifying this!

> This makes the test output a little less pretty, […]

It might also create an incentive to do something about this and implement the currently missing tests (though the tests that are being skipped here are definitely not high priority).

PR #61, opened 2016-08-12
Inconsistent fat file tests

woodruffw:
This PR will contain unit tests for so-called "inconsistent" fat files (i.e., universal Mach-Os whose slices reference different dylibs).

Relevant notes:
* `lipo` allows users to create completely nonsensical fat files, like the combination of a ppc64 executable and an x86_64 bundle. I can't think of any real use case for this sort of behavior, so we should probably consider those types of files simply broken (and not give them special treatment).

cc @UniqMartin


test/src/Makefile
```diff
@@ -110,6 +110,11 @@ $(ALL_DIRS):
 %/hellobundle.so: $(LIBHELLO_SRC) %
 	$(CC) $(ARCH_FLAGS) -bundle $< -o $@
 
+# Run post-make builds (inconsistent binaries)
+.PHONY: post-make
+post-make:
+	bash post-make.sh $(USE_DIRS)
```
UniqMartin: How do you feel about making `post-make.sh` executable and then executing it directly via `./post-make.sh` in this line? It feels a bit strange to have a shebang in that file and then to ignore it.

test/src/post-make.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+
+function main() {
```
UniqMartin: This is an interesting style for shell functions I haven't seen anywhere else. The two styles I've seen are:

- `function main {` (note the lack of parentheses)
- `main() {`

Among these, I heavily prefer the latter. It seems to be the most widespread one in my experience and it is consistent with what we do in Homebrew code. What do you think?

On a side note: Would you mind converting this to a two-space indent with spaces, like we tend to favor in all other Homebrew code I'm aware of (with `Makefile`s being an obvious exception)? You might want to tweak the `.editorconfig` for that.

test/src/post-make.sh
```diff
@@ -0,0 +1,39 @@
...
+	for use_dir in "${@}"; do
```
UniqMartin: Best practice would be to declare `use_dir` as a local variable.

test/src/post-make.sh
```diff
@@ -0,0 +1,39 @@
...
+		# we only want to build libinconsistent.dylib as a fat mach-o
+		if [[ "${use_dir}" == fat* ]]; then
```
UniqMartin: Best practice is to favor `=` over `==` when doing string comparison in shell code. Another thing you might want to consider that avoids too much nesting is the shell equivalent of `next if` in Ruby, e.g.:

```shell
[[ "${use_dir}" = fat* ]] || continue
inconsistent_for "${use_dir}"
```

test/src/post-make.sh
```diff
@@ -0,0 +1,39 @@
...
+			inconsistent_for "${use_dir}"
+		fi
+	done
+}
+
+function inconsistent_for() {
+	# this mess just splits a fat directory spec like fat-i386-x86_64 into
+	# its constituent arch(3) pairs (e.g., i386 and x86_64)
+	local fat_dir="${1}"
+	local split_fat_dir=(${fat_dir//-/ })
```
UniqMartin: If you think this is a mess (as I'm inferring from your comment above), here's an alternative:

```shell
IFS=- read -a split_fat_dir <<<"${fat_dir}"
```

You can even strip the `fat-` prefix beforehand, if you like:

```shell
IFS=- read -a split_fat_dir <<<"${fat_dir#fat-}"
```

test/src/Makefile
```diff
@@ -110,6 +110,11 @@ $(ALL_DIRS):
 %/hellobundle.so: $(LIBHELLO_SRC) %
 	$(CC) $(ARCH_FLAGS) -bundle $< -o $@
 
+# Run post-make builds (inconsistent binaries)
+.PHONY: post-make
+post-make:
```
UniqMartin: Best practice would be to declare the dependencies of this make target nonetheless, i.e.:

```makefile
post-make: $(USE_TARGETS)
```

This way, a `make post-make` (maybe name the target `inconsistent` and the shell script `make-inconsisten.sh`?) will simply work instead of terminating as follows (though I give you credit for catching this in the shell script):

```console
$ make post-make
Makefile:49: USE - Option either unset or invalid. Using a safe fallback.
Makefile:50: USE - Valid choices: all, 10.6-xcode3.2.6, 10.11-xcode7.3.
bash post-make.sh i386 x86_64 fat-i386-x86_64
Error: Missing file: i386/libhello.dylib. Did you run make?
make: *** [post-make] Error 1
```

UniqMartin:
I've left a few mostly stylistic comments (and most of them regarding the shell script), but overall I'm very pleased with this first take. :+1: Also the test reads really nicely with just the right balance of guiding comments and assertions. 💯

> Relevant notes:
> * `lipo` allows users to create completely nonsensical fat files, like the combination of a ppc64 executable and an x86_64 bundle. I can't think of any real use case for this sort of behavior, so we should probably consider those types of files simply broken (and not give them special treatment).

I never checked this thoroughly, but I'm not entirely surprised. And I agree, that's pretty silly.

test/src/Makefile
woodruffw: Sounds good to me!

test/src/post-make.sh
woodruffw: The latter is fine by me! I'll also tweak the `.editorconfig`.

test/src/post-make.sh
woodruffw: I was being mostly sarcastic (since I always end up looking up the syntax for string splitting/truncation in bash), but the second alternative looks great to me :smile:

test/src/make-inconsistent.sh
```diff
@@ -0,0 +1,41 @@
...
+  local lib2="${arch2}/libextrahello.dylib"
+
+  [[ -f "${lib1}" ]] || die "Missing file: ${lib1}. Did you run make?"
+  [[ -f "${lib2}" ]] || die "Missing file: ${lib2}. Did you run make?"
+
+  echo "[+] Creating libinconsistent.dylib for ${fat_dir}"
+  lipo -create "${lib1}" "${lib2}" -output "${fat_dir}/libinconsistent.dylib"
+}
+
+function die() {
```
UniqMartin: Final nit: the `function` here.

woodruffw: Oops; sorry!

UniqMartin:
Nice work! :tada:

woodruffw:
Thanks!

PR #62, opened 2016-08-17
FatFile refactoring

woodruffw:
Nothing major here, just getting rid of the `synchronize_` terminology we decided against earlier with `MachOFile` and making `FatFile` more similar to its single-arch counterpart.


UniqMartin:
Again, some nice refactoring and cleanup! :+1: Thanks!

PR #64, opened 2016-09-03
test: Add style enforcement to test script.

woodruffw:
Not much to see here.


woodruffw:
Adding `rubocop` as a test dependency has the unfortunate side effect of making the test system require ruby `>= 2.0.0`.

I don't think we should consider this a major issue, since 1.9.3 has been EOL'd for over a year and Homebrew itself is now vendoring ruby 2 and above.

PR #90, opened 2018-06-05
Add new load commands

mistydemeo:


lib/macho/load_commands.rb
```diff
@@ -141,6 +143,8 @@ module LoadCommands
       :LC_LINKER_OPTIMIZATION_HINT => "LinkeditDataCommand",
       :LC_VERSION_MIN_TVOS => "VersionMinCommand",
       :LC_VERSION_MIN_WATCHOS => "VersionMinCommand",
+      :LC_NOTE => "LoadCommand",
+      :LC_BUILD_VERSION => "VersionMinCommand",
```
woodruffw: This probably needs a new class, based on the corresponding structure in LLVM's Mach-O parser:

```c
    struct build_version_command {
        uint32_t    cmd;        /* LC_BUILD_VERSION */
        uint32_t    cmdsize;    /* sizeof(struct build_version_command) plus */
        /* ntools * sizeof(struct build_tool_version) */
        uint32_t    platform;   /* platform */
        uint32_t    minos;      /* X.Y.Z is encoded in nibbles xxxx.yy.zz */
        uint32_t    sdk;        /* X.Y.Z is encoded in nibbles xxxx.yy.zz */
        uint32_t    ntools;     /* number of tool entries following this */
    };
```

woodruffw: (For comparison, `version_min_command` has just two fields: `version` and `sdk`).

lib/macho/load_commands.rb
```diff
@@ -141,6 +143,8 @@ module LoadCommands
       :LC_LINKER_OPTIMIZATION_HINT => "LinkeditDataCommand",
       :LC_VERSION_MIN_TVOS => "VersionMinCommand",
       :LC_VERSION_MIN_WATCHOS => "VersionMinCommand",
+      :LC_NOTE => "LoadCommand",
```
woodruffw: Based on http://llvm.org/doxygen/Support_2MachO_8h_source.html, this is the corresponding structure:

```c
     struct note_command {
       uint32_t cmd;        // LC_NOTE
       uint32_t cmdsize;    // sizeof(struct note_command)
       char data_owner[16]; // owner name for this LC_NOTE
       uint64_t offset;     // file offset of this data
       uint64_t size;       // length of data region
  };
```

It looks like `LC_NOTE` is only being used in `MH_CORE` files, though, so this doens't need to be implemented in this PR :slightly_smiling_face:

mistydemeo: Yes, I haven't encountered it yet; I just added it since I was adding `0x32`, which I have encountered. 😄

lib/macho/load_commands.rb
mistydemeo: Change made and tested with a binary. No `build_tool_version` yet.

woodruffw: `TwoLevelHintsCommand` might be a good reference for adding `build_tool_version`, since it encapsulates its own table/array structure.

mistydemeo: OK, updated. We weren't handed an offset to read the table from so I'm making the assumption here that it's the offset following `view.offset + 24`.

While we're not given much info about the tool structure, I am seeing what looks like correct information from what we do have:

```ruby
irb(main):005:0> command.platform #=> 1, e.g. macOS
irb(main):006:0> command.minos_string #=> "10.14.0"
irb(main):007:0> command.sdk_string #=> "10.14.0"
irb(main):008:0> command.tool_entries.tools.first.tool #=> 3, e.g. ld
irb(main):009:0> command.tool_entries.tools.first.version #=> 26738945, format of this string not documented
```

woodruffw: > We weren't handed an offset to read the table from so I'm making the assumption here that it's the offset following view.offset + 24.

Yeah, I think this is a reasonable assumption.

lib/macho/load_commands.rb
```diff
@@ -1119,6 +1123,97 @@ def sdk_string
...
+      # @return [Integer]
+      attr_reader :platform
+
+      # @return [Integer] the minimum OS version X.Y.Z packed as x16.y8.z8
+      attr_reader :minos
+
+      # @return [Integer] the SDK version X.Y.Z packed as x16.y8.z8
+      attr_reader :sdk
+
+      # @return ToolEntries tool entries
```
woodruffw: Nit: `ToolEntries` -> `[ToolEntries]` for YARD

woodruffw:
Other than that nit, this looks good to me! Thanks a ton @mistydemeo :smile:

lib/macho/load_commands.rb
mistydemeo: Fixed!

mistydemeo:
Great! Thanks for the review. I'll merge this when CI goes green.

PR #115, opened 2018-09-29
test: Re-enable Ruby 2.0, 2.1

woodruffw:
The main reason for disabling these was that RuboCop stopped
supporting them. By tweaking run-tests to only run RuboCop on
supported Rubies, we can circumvent the problem.


woodruffw:
...or so I thought. It looks as though the present code needs 2.3 or greater, and that rewinding back to earlier versions won't be trivial (since the current code is frozen-string-literal-correct and makes heavy use of the merged `Fixnum`/`Integer` class).

woodruffw:
Closing due to infeasibility.

PR #265, opened 2020-10-06
WIP: Code signing data parsing

### LINKED ISSUE #262: Pure-Ruby Code Directory/Code signing parsing and manipulation
I'm going to use this issue as a dumping ground as I explore a pure-Ruby alternative to #260.

At a high level: 

* If a binary already contains an `LC_CODE_SIGNATURE`, we need to erase it and replace it with our own (ad-hoc) signature
* If a binary doesn't contain an `LC_CODE_SIGNATURE`, we need to add a new load command containing one

That's not the end of things:

* `LC_CODE_SIGNATURE` *references* the signing data, but doesn't actually contain it. It's actually hiding in the `__LINKEDIT` segment. That means that we'll need to rewrite (and probably resize) `__LINKEDIT`.

woodruffw:
In progress.

Closes #262.


rickmark:
So I think we should move from `LinkeditDataCommand` to the idea of a meta-programming model where the various LC_ commands are mapped too a class (LC_CODE_SIGNING => CSSuperBlob).  Today the fact that it groups unrelated things feels off.

woodruffw:
Could you clarify what unrelated things are currently being grouped? The current mapping is from each load command’s ID to the parsed form, with a generic `LoadCommand` as fallback for when we don’t know how to parse an ID.

PR #283, opened 2020-11-06
fix(macho): Silence the call to `codesign`

woodruffw:
Replaces `system` with `Open3.capture3` to collect `codesign`'s output and error streams.


mistydemeo (approved):
Confirmed this works, silently, while still raising `MachO::CodeSigningError` when signing fails.

PR #370, opened 2021-05-30
macho: updates from cctools-973.0.1

woodruffw:
WIP. I need to implement `fileset_entry_command`:

```c
/*
 * LC_FILESET_ENTRY commands describe constituent Mach-O files that are part
 * of a fileset. In one implementation, entries are dylibs with individual
 * mach headers and repositionable text and data segments. Each entry is
 * further described by its own mach header.
 */
struct fileset_entry_command {
    uint32_t     cmd;        /* LC_FILESET_ENTRY */
    uint32_t     cmdsize;    /* includes entry_id string */
    uint64_t     vmaddr;     /* memory address of the entry */
    uint64_t     fileoff;    /* file offset of the entry */
    union lc_str entry_id;   /* contained entry id */
    uint32_t     reserved;   /* reserved */
};
```


woodruffw:
Also needs unit tests. I'm pretty sure filesets are part of the new kernelcache format, so I can probably just steal one of those.

Refs:

* https://hex-rays.com/products/ida/news/7_5sp2/
* https://github.com/NationalSecurityAgency/ghidra/issues/2487

woodruffw:
Looks like `ruby-macho` already chokes on kernelcache Mach-Os, since we don't support Apple's sneaky compressed header. We also probably won't support decompressing any time soon, since that would require us to vendor (or implement) LZSS and LVZN support.

woodruffw:
The prelinked Mach-O header format shows up in `kernelcache.h` in `kext_tools`. I've annotated it a bit below:

```c
// prelinkVersion value >= 1 means KASLR supported
typedef struct prelinked_kernel_header {
    uint32_t  signature;  // WW: "COMP"
    uint32_t  compressType; WW: "LZSS" | "LZVN"
    uint32_t  adler32; // WW: crc32?
    uint32_t  uncompressedSize;
    uint32_t  compressedSize;
    uint32_t  prelinkVersion;
    uint32_t  reserved[10];
    char      platformName[PLATFORM_NAME_LEN]; // unused
    char      rootPath[ROOT_PATH_LEN];         // unused
    char      data[0];
} PrelinkedKernelHeader;
```

woodruffw:
Initial Ruby support for LZVN: https://github.com/woodruffw/lzfse.rb

woodruffw:
This now works, more or less:

```bash
$ gem install lzfse
# in the ruby-macho repo
$ irb -Ilib
```

```ruby
>> require 'macho'
>> f = MachO::FatFile.new("/path/to/some/prelinkedkernel", decompress: true)
```

The `prelinked` kernel on my (x86-64) macOS host is inexplicably wrapped in a fat header, but I have no idea if that's normal.

woodruffw:
I'm not sure how best to test this, since (AFAICT) there are no real examples of this load command or compressed Mach-Os outside of Apple's prelinked kernel binaries. And it's probably a bad idea to distribute those in the unit tests.

PR #426, opened 2022-01-11
Update gitignore for common IDEs

rickmark:
Makes working in common IDEs easier by ignoring IDE specific files

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,10 +5,11 @@ on:
   pull_request: []
 jobs:
   tests:
-    runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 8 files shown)


rickmark:
heh race condition

woodruffw:
I don't mind merging this, but out of curiosity: are you using a global `.gitignore`? That's where I usually put common IDE cruft, so that I don't have to copy it into each project 🙂

rickmark:
Problem is every potential contributor won’t be - it’s a common PR I end up doing to make up streaming easier for new people

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

woodruffw (approved):

PR #431, opened 2022-01-13
Fix for using `macho` in Pry/IRB

rickmark:
Using 'macho' in Pry/IRB was slow due to the editor printing the entire contents of @raw_data, this makes the inspect element print only the length of the view.

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -1839,6 +1839,11 @@ module MachO
           "reserved" => reserved,
         }.merge super
       end
+
+      # @return [SegmentCommand64, nil] the matching segment command or nil if nothing matches
+      def segment
+        view.macho_file.command(:LC_SEGMENT_64).select { |cmd| cmd.fileoff == fileoff }.first
+      end
     end
   end
 end
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -592,7 +592,7 @@ module MachO
           LoadCommands::LoadCommand
         end
 
-        view = MachOView.new(@raw_data, endianness, offset)
+        view = MachOView.new(self, @raw_data, endianness, offset)
         command = klass.new_from_bin(view)
 
         load_commands << command
--- a/lib/macho/view.rb
+++ b/lib/macho/view.rb
@@ -29,5 +29,9 @@ module MachO
         "offset" => offset,
       }
     end
+
+    def inspect
+      "#<#{self.class}:0x#{self.__id__.to_s(16)} @endianness=#{@endianness.inspect}, @offset=#{@offset.inspect}, length=#{@raw_data.length}>"
+    end
   end
 end
--- a/lib/macho/view.rb
+++ b/lib/macho/view.rb
@@ -3,6 +3,9 @@
 module MachO
   # A representation of some unspecified Mach-O data.
   class MachOView
+    # @return [MachOFile] that this view belongs to
+    attr_reader :macho_file
+
     # @return [String] the raw Mach-O data
     attr_reader :raw_data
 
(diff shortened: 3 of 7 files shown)


woodruffw:
LGTM! Mind fixing the lint?

```
Offenses:

lib/macho/view.rb:34:28: C: [Correctable] Style/RedundantSelf: Redundant self detected.
      "#<#{self.class}:0x#{self.__id__.to_s(16)} @endianness=#{@endianness.inspect}, @offset=#{@offset.inspect}, length=#{@raw_data.length}>"
                           ^^^^

11 files inspected, 1 offense detected, 1 offense auto-correctable
Error: Process completed with exit code 1.
```

rickmark:
Doing so now - also adding a MachOFile reference to MachOView so that you can navigate backwards from a command to the file (some commands make use of other commands in a file such as MH_FILESET_ENTRY to the LC_LOAD at the same address)

woodruffw:
> also adding a MachOFile reference to MachOView so that you can navigate backwards from a command to the file (some commands make use of other commands in a file such as MH_FILESET_ENTRY to the LC_LOAD at the same address)

Cool, that should be fine as long as we don't break any APIs that allow for dynamic construction/creation of LCs 🙂

rickmark:
Looking to exercise new code paths now

rickmark:
> > also adding a MachOFile reference to MachOView so that you can navigate backwards from a command to the file (some commands make use of other commands in a file such as MH_FILESET_ENTRY to the LC_LOAD at the same address)
> 
> Cool, that should be fine as long as we don't break any APIs that allow for dynamic construction/creation of LCs 🙂

Simple solution was to attach the MachOFile to the MachOView so navigation would be `command -> view -> macho_file -> command` which could be shortened by delegation but hey

rickmark:
Ok I'm a bad person... This PR expanded scope to cover the use case.

The E2E capability is load BootKernelCollection in pry (fast due to the @raw_data fix)

Iterate over FH_FILESET_ENTRY and map each to .segment and extract

rickmark:
Added some tests to make CI happy

woodruffw:
LGTM structurally, although I'm a little bit confused by what's going on with `FilesetEntryCommand#segment` here -- it looks like you're using it to find a corresponding segment load command for the embedded Mach-O, but I was under the impression that `fileoff` could be *within* a segment and not just at the start of one. Is that a misunderstanding on my part?

rickmark:
> LGTM structurally, although I'm a little bit confused by what's going on with `FilesetEntryCommand#segment` here -- it looks like you're using it to find a corresponding segment load command for the embedded Mach-O, but I was under the impression that `fileoff` could be _within_ a segment and not just at the start of one. Is that a misunderstanding on my part?

Let me do some more testing - it seems to me that the .kc file has `__REGION{digit}` that have a file offset - the FILESET_INFO has a matching fileoff and annotates

I'm thinking of creating a Git LFS repo for text mach-os (large ones like kc files and prelink kernels that can be brought into testing if you want the other larger types.

woodruffw:
Makes sense! I didn't realize each had a `__REGION`.  If that's the observed behavior (knowing Apple, I'm sure it isn't documented), then I'm good with this approach.

rickmark:
Have you considered adding “overcommit” to stop my being offensive?  Peer of mine at Coinbase made it

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

rickmark:
Oh it’s commit hooks for developers to prevent getting PRs that require simple changes like Rubocop.  Clint side only not server

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

woodruffw:
Oh, I misunderstood. Sure, as long as they're opt-in, I would accept a PR enabling them for `rubocop`.

woodruffw (approved):
LGTM! We should give some though to perhaps refactoring `MachOView` entirely so that it doesn't effectively have two copies/references of the raw Mach-O data in it, now that we store both `macho_file` and `raw_data`. In particular, `raw_data` is probably replaceable with an `offset` and `length` for `macho_file.raw_data`. But that can wait for another PR.

Issue #72, opened 2017-07-04
Better stringification for load commands.


woodruffw:
Right now, every load command (e.g., `MachO::LoadCommand::DylibCommand` just returns its type when stringified:

```ruby
dylib.to_s # => "LC_DYLIB_ID"
```

This is a good default for load commands that don't have a nice string representation, but we can do better for dylib commands, rpaths, versions, etc:

```ruby
dylib.to_s # => "/usr/lib/whatever.dylib"
rpath.to_s # => "/var/lib/foo"
uuid.to_s # => "4ecc2bd4-4c76-4c94-98b3-a650f0e5af17"
```

This will also eliminate the need to do `dylib.name.to_s` and similar chains when extracting strings from `LCStr` instances within load commands.

This changes the current stringification behavior substantially (even if it doesn't break the CI), so it's probably best for a 2.x release.

woodruffw:
#447

PR #447, opened 2022-03-07
Improved stringification for load commands

apainintheneck:
This PR is related to issue #72.

I added string representations for all of the LoadCommand classes where the .to_s method would return the only obvious, expected result. In some cases, that means that the .to_s just acts as an alias for another method and in other cases it meant just calling the LCStr.to_s method. Either way I didn't delete any of the old methods so there should be no problems with backwards compatibility.

### What Changed
1. Changes to load_commands.rb
	Added .to_s to all easily stringifiable LoadCommand classes
2. Changes to macho_file.rb
	Removed now unnecessary chained method calls to LoadCommand classes

### DIFF
--- a/lib/macho/macho_file.rb
+++ b/lib/macho/macho_file.rb
@@ -302,7 +302,7 @@ module MachO
 
       dylib_id_cmd = command(:LC_ID_DYLIB).first
 
-      dylib_id_cmd.to_s
+      dylib_id_cmd.name.to_s
     end
 
     # Changes the Mach-O's dylib ID to `new_id`. Does nothing if not a dylib.
@@ -368,7 +368,7 @@ module MachO
     # All runtime paths searched by the dynamic linker for the Mach-O.
     # @return [Array<String>] an array of all runtime paths
     def rpaths
-      command(:LC_RPATH).map(&:to_s)
+      command(:LC_RPATH).map(&:path).map(&:to_s)
     end
 
     # Changes the runtime path `old_path` to `new_path`
@@ -382,7 +382,7 @@ module MachO
     # @return [void]
     # @raise [RpathUnknownError] if no such old runtime path exists
     def change_rpath(old_path, new_path, options = {})
-      old_lc = command(:LC_RPATH).find { |r| r.to_s == old_path }
+      old_lc = command(:LC_RPATH).find { |r| r.path.to_s == old_path }
       raise RpathUnknownError, old_path if old_lc.nil?
 
       new_lc = LoadCommands::LoadCommand.create(:LC_RPATH, new_path)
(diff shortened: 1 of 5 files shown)


lib/macho/macho_file.rb
```diff
@@ -302,7 +302,7 @@ def dylib_id
 
       dylib_id_cmd = command(:LC_ID_DYLIB).first
 
-      dylib_id_cmd.name.to_s
```
woodruffw, line 305: Let's preserve the original API uses here -- `path`, `name`, etc. are more semantic and not context-dependent the way `to_s` is.

lib/macho/macho_file.rb
```diff
@@ -368,7 +368,7 @@ def change_install_name(old_name, new_name, _options = {})
     # All runtime paths searched by the dynamic linker for the Mach-O.
     # @return [Array<String>] an array of all runtime paths
     def rpaths
-      command(:LC_RPATH).map(&:path).map(&:to_s)
```
woodruffw, line 371: Same here and below.

woodruffw:
Overall LGTM. I'd prefer we keep the uses in `macho_file.rb` the way they are, but the `#to_s` implementations themselves look good!

apainintheneck:
Ah, that makes sense. I just reverted that file to its original state.

woodruffw (approved):

woodruffw:
LGTM. Could you add a few small tests for the new `#to_s` impls?

apainintheneck:
I was able to add to the tests in the test_create_load_commands.rb file for the DyLibCommand and RpathCommand classes. Most of the other LoadCommand classes don't really seem to have tests yet as far as I can see so I don't really know where to add those tests.

When I added an assert to the test_macho.rb file related to the SegmentCommand class it brought up a few more linter errors which I solved by changing the style and excluding one of the cops from the test folder. These linter errors only showed up on the pre-commit checks not when running the tests beforehand which was weird.

apainintheneck:
Never mind I now see that run-tests doesn't actually run the linter so those warnings were always there. I think that it still might be a good idea to exclude that cop from the tests folder either way.

It spits out the following warning type three times.

```
test/test_macho.rb:150:11: W: Lint/AmbiguousBlockAssociation: 
Parenthesize the param MachO::Sections::SECTION_ATTRIBUTES.keys.any? { |sa| sect.attribute?(sa) } 
to make sure that the block will be associated with the MachO::Sections::SECTION_ATTRIBUTES.keys.any? method call.
          assert MachO::Sections::SECTION_ATTRIBUTES.keys.any? { |sa| sect.attribute?(sa) }
```

I'm gonna go over to the Rubocop repo and see what they have to say about it.

woodruffw:
LGTM, thanks.

Issue #139, opened 2019-03-15
Fat parser should fail on file with 0 architectures


woodruffw:
Currently, ruby-macho will happily parse a fat Mach-O with no internal slices (i.e., `nfat_arch == 0`). This should be an error (potentially an ignorable one with `permissive: true`).

Reference this LLVM object: https://github.com/llvm-mirror/llvm/blob/master/test/Object/Inputs/macho-invalid-fat-header

Observed behavior:

```ruby
>> MachO.open('macho-invalid-fat-header')
#<MachO::FatFile:0x000055776df6b2a8 @filename="macho-invalid-fat-header", @options={}, @raw_data="\xCA\xFE\xBA\xBE\x00\x00\x00\x00", @header=#<MachO::Headers::FatHeader:0x000055776df6b118 @magic=3405691582, @nfat_arch=0>, @fat_archs=[], @machos=[]>
```

Expected behavior:

An exception.

woodruffw:
Fixed with #449 a long time ago.

PR #449, opened 2022-03-19
Fat parser should fail on file with 0 architectures

apainintheneck:
This PR is related to issue #139.

I added a new exception called ZeroArchitectureError that gets thrown every time the permissive option is not set and nfat_arch == 0.

The only question I have is how you'd like to test for this exception. I just grabbed the llvm binary that you linked to in the original post and threw it in the test/bin/llvm/ folder and used that for testing. Is that what you were thinking? Also, that probably means that we should include some sort of llvm license in that directory.


lib/macho/fat_file.rb
```diff
@@ -346,6 +348,11 @@ def populate_fat_header
       # formats.
       raise JavaClassFileError if fh.nfat_arch > 30
 
+      # Rationale: fat Mach-0 files with no internal slices should
+      # return an error unless the permissive option is set.
+      permissive = options.fetch(:permissive, false)
```
woodruffw, line 353: IMO we can remove the permissive check here: there's very little a user can do with an empty universal binary, so we should fail unconditionally if we have no architecture slices.

woodruffw:
>  Is that what you were thinking? Also, that probably means that we should include some sort of llvm license in that directory.

Yep, that's what I had in mind. Re: licensing: it should be sufficient to put a license notice in the top-level README, noting that we use binaries taking from the LLVM project for the test suite.

apainintheneck:
Copy and pasting the entire license would be a bit much for the top level README though, wouldn't it? Their Apache license is significantly longer than the current README. How about I just drop it in the llvm folder and then link to it from the README?

lib/macho/fat_file.rb
apainintheneck: That makes sense.

woodruffw:
> How about I just drop it in the llvm folder and then link to it from the README?

Yep, that works.

woodruffw (approved):
LGTM modulo adding the license notice. Thanks!

apainintheneck:
Okay, I added the license notice.

woodruffw (approved):

Issue #140, opened 2019-03-15
Fat parser should fail on mismatched CPU types


woodruffw:
Currently, ruby-macho will happily parse a fat Mach-O whose `fat_archs` and internal slices have mismatching (i.e., not 1-to-1) CPU types and subtypes.

Reference this LLVM object: https://github.com/llvm-mirror/llvm/blob/master/test/Object/Inputs/macho-invalid-fat_cputype

Observed behavior:

```ruby
>> MachO.open('macho-invalid-fat_cputype')
#<MachO::FatFile:0x000055776df7b0e0 @filename="macho-invalid-fat_cputype", @options={}, @raw_data="\xCA\xFE\xBA\xBE\x00\x00\x00\x01\x00\x00\x00\f\x00\x00\x00\x00\x00\x00\x00\x1C\x00\x00\x00\x1C\x00\x00\x00\x02\xCE\xFA\xED\xFE\a\x00\x00\x00\x03\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00", @header=#<MachO::Headers::FatHeader:0x000055776df7af50 @magic=3405691582, @nfat_arch=1>, @fat_archs=[#<MachO::Headers::FatArch:0x000055776df7ae10 @cputype=12, @cpusubtype=0, @offset=28, @size=28, @align=2>], @machos=[#<MachO::MachOFile:0x000055776df7ad20 @filename=nil, @options={}, @raw_data="\xCE\xFA\xED\xFE\a\x00\x00\x00\x03\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00", @endianness=:little, @header=#<MachO::Headers::MachHeader:0x000055776df7ab18 @magic=4277009102, @cputype=7, @cpusubtype=3, @filetype=1, @ncmds=0, @sizeofcmds=0, @flags=0>, @load_commands=[]>]>
```

Expected behavior:

An exception.

woodruffw:
Fixed with #450 a long time ago.

PR #450, opened 2022-03-21
Added CPUTypeMismatchException

apainintheneck:
Okay, this is pretty similar to the last one. It's linked to issue #140.

1. New exception triggers when there is a cputype or cpusubtype
mismatch between fat header and internal macho slice in the
FatFile#populate_machos method.

2. Added a test for it along with adding a new binary file
to the test/bin/llvm folder.


woodruffw (approved):
LGTM, thanks!

PR #454, opened 2022-03-31
New MachOStructure DSL

### LINKED ISSUE #70: New DSL for MachOStructure
This is just for collecting my thoughts and getting feedback on a DSL for `MachOStructure` (and, by inheritance, just about every Mach-O data-oriented class in `ruby-macho`.

The current "DSL" for a Mach-O structure looks something like this: 

```ruby
class SomeImportantStruct < MachOStructure
  attr_reader :something
  attr_reader :something_else

  FORMAT = "L=2" # something and something_else are "L", meaning int32s
  SIZEOF = 8 # 2 x sizeof(int32)

  def initialize(something, something_else)
    # ...
  end 
end
```

This is incredibly repetitive, bug-prone, and difficult to read (there's no immediate way to determine the size of each field).

Something like this would be *substantially* better:

```ruby
class SomeImportantStruct < MachOStructure
  field :int32, :something
  field :int32, :something_else

  # FORMAT, SIZEOF, and instantiation are automatically generated from the field definitions
end
```

Extremely early versions of `ruby-macho` actually had something like this, using `CStruct` [from this repo](https://github.com/samsonjs/compiler), but I dropped it for reasons that I don't remember (probably having to do with me being unfamiliar with ruby DSL writing at the time).

This will probably require a large amount of refactoring and breakage, so I don't think it'll make it into 1.1 (or even 1.x, it might be destined for a major bump).

apainintheneck:
Here's my first attempt at a new dsl for the MachOStructure class. At the moment, it seems to be functional but not nearly as performant as the previous hardcoded version. That seems to be related to all logic stuffed in the constructor to deal with all of the edge cases needed to initialize derived classes. So I think the next step will be to move some of that out of the constructor and into the #field definition method. Of course, it's Ruby so that work won't happen at compile time but I'm hoping it will help speed it up a little.

Other than that there is also the possibility of generating the #serialize, #to_s and #to_h methods as well which I haven't explored yet. Let me know what you think of it so far.

Resolves #70


woodruffw:
Thanks! I'll try and get to this tomorrow, or over the weekend.

apainintheneck:
I managed to pull essentially all of the logic out of the constructor and moved it into the #field method and a bunch of helpers that are used to define new methods. This makes it easier to hide meta programming methods away and the performance now seems comparable to the master branch. Also, each variable reader method now evaluates lazily which is kind of necessary since everything was moved out of the constructor.

apainintheneck:
One thing I'm not really sure about though is how to go about adding tests for the new DSL since most of the new methods are used to just generate methods.

lib/macho/headers.rb
```diff
@@ -505,30 +505,14 @@ module Headers
...
-      # @api private
-      SIZEOF = 8
-
-      # @api private
-      def initialize(magic, nfat_arch)
-        super()
-        @magic = magic
-        @nfat_arch = nfat_arch
-      end
+      field :nfat_arch, :uint32_net
```
woodruffw, line 511: This is a very minor thing, but instead of having a separate `:uint32_net` specifier we should try something like:

```ruby
field :nfat_arch, :uint32, :endian => :big
```

Where `:endian` is not specified by default for fields that have their endianness determined by the overarching container.

apainintheneck: Oh yeah, that would look nicer. I'll see what I can do.

lib/macho/headers.rb
```diff
@@ -548,42 +532,23 @@ def to_h
     # @see MachO::Headers::FatHeader
     class FatArch < MachOStructure
       # @return [Integer] the CPU type of the Mach-O
-      attr_reader :cputype
+      field :cputype, :uint32, :endian => :big
 
       # @return [Integer] the CPU subtype of the Mach-O
-      attr_reader :cpusubtype
+      field :cpusubtype, :uint32, :endian => :big, :mask => CPU_SUBTYPE_MASK
```
woodruffw, line 538: Nice!

lib/macho/headers.rb
```diff
@@ -815,54 +733,31 @@ def to_h
...
-      attr_reader :compressed_size
+      field :compressed_size, :uint32, :endian => :big
 
       # @return [Integer] the version of the prelink format
-      attr_reader :prelink_version
+      field :prelink_version, :uint32, :endian => :big
 
       # @return [void]
-      attr_reader :reserved
+      field :reserved, :binary, :size => 40, :unpack => "L>10"
```
woodruffw, line 755: The name `:binary` might be a little confusing here, since `a{N}` is really a NUL-padded string of up to `{N}` bytes.

Spitballing:

```ruby
field :reserved, :strz, :pad => 40
```

...to emphasize that it's NUL-terminated and that the size is padding.

apainintheneck: It really is annoying that we don't have one word to describe a null terminated binary slice in English. Originally I had it as `:bin_string` which I think we can both agree is worse and `:binary` seemed like an improvement but I'm not really attached to it or anything. I'll sleep on it and see what I come up with in the morning.

Also, I think that adding `:pad` as an alias for `:size` is a trivial change that would improve the readability of the code a bit. It feels more declarative. Functionally they'd work the same way, right?

apainintheneck: Okay, after brainstorming a little bit more here are a few mediocre options. Also, I'm not feeling `:pad` anymore. It seems unnecessary imo.

```Ruby
field :reserved, :strz, :size => 40
field :reserved, :var, :size => 40                                              
field :reserved, :varchunk, :size => 40                                         
field :reserved, :varbyte, :size => 40                                          
field :reserved, :nullslice, :size => 40                                        
field :reserved, :nchunk, :size => 40                                           
field :reserved, :nullchunk, :size => 40
```

apainintheneck: Apparently, `varbyte` already exists as a name to refer to exactly this sort of thing though I doubt it's super well known outside of some database versions. But if we decide to use that then maybe it would be nice to use `varchar` for symmetry instead of `string`. Either way it's your call.

woodruffw: The second hardest problem in CS strikes again :slightly_smiling_face: 

I also gave this some more thought, and realized that we might want to decompose this the way we did the `_net` format earlier. Something like this, perhaps:

```ruby
field :foo, :string, :padding => :null, :size => 40
```

or even just have `:padding` be the literal padding byte itself, e.g. `:padding => "\x00"` (although this is harder to convert back into a `#pack`-style format).

This has a few advantages:

* We're separating parsing and semantics: the thing we're extracting is fundamentally a string, not a "strz" or anything else. Ruby calls it a string, so we should call it a string too.
* It gives us the ability to handle alternate paddings, like space paddings. We don't currently need these in Mach-O (AFAIK), but they'd fit naturally in the future.

Thoughts?

apainintheneck: Kind of unrelated but I just realized that the unpacking directive that's currently used for binary strings doesn't remove trailing nulls and spaces. It just grabs the whole chunk. That's the difference between 'a' (what we're currently using) and 'A'. How confusing!

Following your lead, would we just assume that the basic `:string` type without any options has no padding? I don't hate this idea and it does leave room in the future for other types of padding (though the `String#unpack` method doesn't really have that many options). I prefer `:null` over `\x00` though.
```ruby
field :bstring, :string, :size => 40
field :cstring, :string, :padding => :null, :size => 40
```

apainintheneck:
Okay, I changed the name of the string types to what you suggested and I like how that turned out. I also added documentation and tests (though that was a few commits ago). I think overall this is in pretty good shape. Let me know if there is anything else I need to do before this gets merged in.

I've had a lot of fun working on this and I've learned a lot so far. This has been a great experience.  😄

woodruffw:
> Okay, I changed the name of the string types to what you suggested and I like how that turned out. I also added documentation and tests (though that was a few commits ago). I think overall this is in pretty good shape. Let me know if there is anything else I need to do before this gets merged in.

Fantastic! This is looking really great; I'll do a final review pass tonight or tomorrow. 

I'm also glad to hear it's been a great experience! Your help keeping ruby-macho maintained is greatly appreciated 🙂

woodruffw:
Sorry, fell off my stack again. Reviewing this evening.

PR #596, opened 2024-04-25
workflows: pin setup-ruby action

Moisan:
Pin the version of the `setup-ruby` action to full length commit SHA as described in the [security hardening for GitHub Actions guide](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
       uses: actions/checkout@main
 
     - name: Set up Ruby
-      uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@70da3bbf44ac06db1b0547ce2acc9380a5270d1e # v1.175.0
       with:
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @Moisan! Some failing CI here but may be able to remove that job.

woodruffw (approved):
(Failed job looks like a flake, re-running.)

woodruffw:
Great, now everything is flaking on CodeCov 🙃

woodruffw:
I guess we need a CodeCov token now. Fixing in another PR.

woodruffw:
Fixed the CI failure with #597. although this is now broken because third-party (fork) PRs don't get access to `secrets.CODECOV_TOKEN`. I'm tempted to just remove the CodeCov integration altogether and replace it with a small workflow step that just spits out the coverage as a PR comment, since we don't treat coverage as a blocker anyways.

PR #598, opened 2024-04-25
workflows: pin setup-ruby action

woodruffw:
Picked from #596 to fix the 3p PR perm issue.

Closes #596.


MikeMcQuaid:
Thanks @woodruffw!

PR #612, opened 2024-05-20
macho/load_commands: support LC_ATOM_INFO

### LINKED ISSUE #609: Mergable Libraries are unsupported
Hello!

Apple has introduced [mergable libraries](https://developer.apple.com/videos/play/wwdc2023/10268/) with Xcode 15.

As I understand them, they are dynamic frameworks but come with additional metadata allowing them to be statically merged.

It also comes with a new load command, `LC_ATOM_INFO` (code 0x36) https://github.com/llvm/llvm-project/blob/82c5d350d200ccc5365d40eac187b9ec967af727/llvm/include/llvm/BinaryFormat/MachO.def#L80 

This library doesn't understand this code and fails parsing the binary.
I'm not sure what you could to with it or what you could expose, I just thought I'd let you know as it breaks parsing altogether.

Have a nice day,
Arnaud

PS: I noticed this issue via CocoaPods, which still uses version 2.5 of this library. So, I don't think there's a need to rush this.

Bo98:
Fixes #609

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -65,6 +65,7 @@ module MachO
       (LC_REQ_DYLD | 0x33) => :LC_DYLD_EXPORTS_TRIE,
       (LC_REQ_DYLD | 0x34) => :LC_DYLD_CHAINED_FIXUPS,
       (LC_REQ_DYLD | 0x35) => :LC_FILESET_ENTRY,
+      0x36 => :LC_ATOM_INFO,
     }.freeze
 
     # association of symbol representations to load command constants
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

woodruffw (approved):
LGTM! It'd be nice to have a source reference for this new LC, but not required.

Bo98:
In this case I used `cctools`: https://github.com/apple-oss-distributions/cctools/blob/7a5450708479bbff61527d5e0c32a3f7b7e4c1d0/include/mach-o/loader.h#L345.

There's also a change coming in macOS 15 that we'll want to support (but not sure if a compiler can output this yet): https://github.com/apple-oss-distributions/cctools/blob/7a5450708479bbff61527d5e0c32a3f7b7e4c1d0/include/mach-o/loader.h#L708-L727

PR #625, opened 2024-06-10
macho/load_commands: support new macOS 15 dylib use command

Bo98:
macOS 15 adds a new `flags` field at the end of the load command. The presence of this is checked in `dyld` by comparing the `timestamp` and `nameoff`:

https://github.com/apple-oss-distributions/dyld/blob/d552c40cd1de105f0ec95008e0e0c0972de43456/mach_o/Header.cpp#L1175

This requires a bit of a hack in `new_from_bin` to basically load the old command, check it and then load the new command instead if conditions match.


woodruffw:
> The presence of this is checked in `dyld` by comparing the `timestamp` and `nameoff`:

Cursed, thank you Apple 🤦. Reviewing now!

Bo98:
If you'd like an example binary, I can send you some

lib/macho/load_commands.rb
```diff
@@ -528,6 +542,35 @@ class DylibCommand < LoadCommand
       # @return [Integer] the library's compatibility version number
       field :compatibility_version, :uint32
 
+      def self.new_from_bin(view)
```
woodruffw, line 545: Minor: it'd be nice to have a comment here explaining why we override `new_from_bin` -- just a summary of what you put in the PR desc would be great 🙂

woodruffw:
> If you'd like an example binary, I can send you some

Yes please! Checking them in with the other test assets would be IMO ideal.

lib/macho/load_commands.rb
```diff
@@ -195,6 +195,20 @@ module LoadCommands
...
+    # @api private
+    DYLIB_USE_FLAGS = {
+      :DYLIB_USE_WEAK_LINK => 0x1,
+      :DYLIB_USE_REEXPORT => 0x2,
+      :DYLIB_USE_UPWARD => 0x4,
+      :DYLIB_USE_DELAYED_INIT => 0x8,
+    }.freeze
+
+    # the marker used to denote a newer style dylib use command.
+    # the value is the timestamp 24 January 1984 18:12:16
```
woodruffw, line 208: Funny choice -- that's the day the Macintosh was released.

Bo98:
> Yes please! Checking them in with the other test assets would be IMO ideal.

I've checked one in. Not actually linked into any test yet but it's there.

Would you like a test prior to merging and releasing? Not poked around the tests here yet but can have a look.

Bo98:
Actually didn't see the Makefile - let me fix that.

woodruffw:
> Would you like a test prior to merging and releasing? Not poked around the tests here yet but can have a look and see which one might fit it best.

No test before merge necessary, but before release would be ideal -- happy to review in a follow-up if you'd like to get this in first.

Bo98:
Did some tests - let me know if you think I've missed anything.

lib/macho/load_commands.rb
```diff
@@ -111,7 +111,7 @@ module LoadCommands
       # "reserved for internal use only", no public struct
       :LC_PREPAGE => "LoadCommand",
       :LC_DYSYMTAB => "DysymtabCommand",
-      :LC_LOAD_DYLIB => "DylibCommand",
+      :LC_LOAD_DYLIB => "DylibUseCommand",
```
Bo98, line 114: Did this and moved the `new_from_bin` hack to there to be slightly more strict where this applies - it should not apply to `LC_ID_DYLIB` etc.

woodruffw (approved):
LGTM, thanks @Bo98!

woodruffw:
@Bo98 LMK when you want a release -- I can do one whenever.

Bo98:
A new release now would be great so I can use this in Homebrew/brew and get a release out for that too.

woodruffw:
Released with 4.1.0!

Bo98:
Thank you!

PR #627, opened 2024-06-11
workflows: add a release workflow

### LINKED ISSUE #273: Add a RubyGems publishing workflow
I currently release new versions of the ruby-macho gem from my desktop. This isn't ideal, both security wise and in terms of availability for other Homebrew maintainers. So, we should use GitHub Actions to automatically publish releases instead.

Some notes:
* It looks like `gem push` can use `GEM_HOST_API_KEY` in the environment to get a RubyGems API key

woodruffw:
This moves ruby-macho from a "William pushes it from his desktop" release process to a CI-managed one, with Trusted Publishing as the authentication mechanism.

Anybody with the ability to create releases on this repo can now release from it. The workflow itself is currently tied to the `rubygems.org` deployment environment, to which we could add signoff restrictions as well.

NB: I've already configured the Trusted Publisher for ruby-macho on RubyGems:

Closes #273

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,30 @@
+name: Release Gem to rubygems.org
+
+on:
+  release:
+    types:
+      - published
+
+permissions:
+  contents: read
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+      id-token: write
+
+    environment: rubygems.org
+
+    steps:
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@0cde4689ba33c09f1b890c1725572ad96751a3fc # v1.178.0
+        with:
+          bundler-cache: true
+          ruby-version: ruby
+
+      # Release
+      - uses: rubygems/release-gem@612653d273a73bdae1df8453e090060bb4db5f31 # v1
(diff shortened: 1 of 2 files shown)


woodruffw:
(Tagged a couple of people for visibility, since this changes a release process.)

p-linnane (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+    steps:
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4.1.6
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@0cde4689ba33c09f1b890c1725572ad96751a3fc # v1.178.0
+        with:
+          bundler-cache: true
+          ruby-version: ruby
+
+      # Release
+      - uses: rubygems/release-gem@612653d273a73bdae1df8453e090060bb4db5f31 # v1
```
Bo98, line 30: I've added this to the actions whitelist!

Looks like it uses `rake release` internally so may need a `Bundler::GemHelper.install_tasks` or whatever in the Rakefile.

woodruffw: Hmm -- I adapted this from `rubygems-await`, which doesn't seem to use that in their Rakefile: https://github.com/segiddins/rubygems-await/blob/main/Rakefile

(But I'm fully cargo-culting this, so maybe I missed an equivalent.)

Bo98: `require "bundler/gem_tasks"` is the magic line that does it for you:

https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/gem_tasks.rb#L7

MikeMcQuaid:
Makes sense, thanks for this @woodruffw!

PR #642, opened 2024-07-01
Drop Ruby 2.6

p-linnane:
Rubocop >= 1.51.0 requires Ruby >= 2.7. Dependabot is throwing errors due to this mismatch.

Since we've officially dropped macOS system Ruby, and 2.6 has been EOL for over 2 years, I think this should be safe.


MikeMcQuaid (approved):
Makes sense to me, thanks @p-linnane!

PR #846, opened 2025-04-24
macho/load_commands: support new load commands from macOS 15.4

Bo98:
Two things to note:

* I haven't tested any of this whatsoever as I do not have a real binary with these commands.
* It's possible some of these are not yet actually emitted by the linker. This was the case last year when `DylibUseCommand` foundations were first found in macOS 14.4 dyld code but not actually enabled until macOS 15.


woodruffw:
Thanks! These look fine to me structurally, although I similarly don't have a test binary. 

@Bo98 do you happen to have source refs for these? I was looking through through the latest https://github.com/apple-oss-distributions/cctools but couldn't find them listed in `loader.h`.

Bo98:
Full dyld and cctools source isn't available yet for macOS 15.4 but I did find `/Library/Developer/CommandLineTools/SDKs/MacOSX15.4.sdk/usr/include/mach-o/loader.h`.

Doesn't tell use whether these are in use yet or what OS versions they apply to like the full source code would but in the end that doesn't really matter much for us.

woodruffw (approved):
Thanks @Bo98! I'll merge here and we can figure out tests if/when these LCs become common.

PR #997, opened 2026-07-10
Add pure-Ruby ad-hoc signing

### LINKED ISSUE #262: Pure-Ruby Code Directory/Code signing parsing and manipulation

MikeMcQuaid:
- remove the `/usr/bin/codesign` dependency from `MachO.codesign!` so Homebrew can repair bottle signatures on Apple Silicon using Ruby
- parse and build XNU-compatible `SuperBlob` and `CodeDirectory` data so replacement signatures can retain security-relevant metadata
- add or replace `LC_CODE_SIGNATURE`, resize `__LINKEDIT` and rebuild fat offsets so every page hash describes the final on-disk layout
- use SHA-256 with legacy SHA-1 agility only for old deployment targets because XNU accepts the open `ld64` ad-hoc format without an identity
- preserve non-linker requirements, entitlements, flags and runtime metadata while deliberately replacing linker-generated metadata
- validate before an in-place write to retain hard links and leave failed inputs unchanged
- cover thin, fat, byte-order, metadata and macOS verification paths with regression tests, and document the compatibility boundaries

Fixes https://github.com/Homebrew/ruby-macho/issues/262

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,30 +56,6 @@ the required ad-hoc signature in Ruby instead of invoking `/usr/bin/codesign`:
 MachO.codesign!("/path/to/my/binary")
 ```
 
-The implementation follows the public structures used by
-[XNU](https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/kern/cs_blobs.h)
-and [ld64](https://github.com/apple-oss-distributions/ld64). For each thin
-Mach-O slice it adds or replaces `LC_CODE_SIGNATURE`, resizes `__LINKEDIT` then
-hashes the final pre-signature bytes in 4 KiB pages. It emits a SHA-256
-CodeDirectory, adding a SHA-1 alternate only when the declared deployment target
-requires legacy hash agility. Fat binaries are signed one slice at a time then
-laid out again with updated architecture offsets and sizes.
-
-When replacing a non-linker signature, the signer preserves its requirements,
-entitlements, flags, runtime version and executable-segment flags. Linker
-signatures use fresh ad-hoc metadata, matching Apple's replacement behaviour.
-Code-signing blobs use their mandated big-endian representation independently
-of the Mach-O byte order, while the load command retains the slice byte order.
-
-The complete signature is built and validated before the file is written. This
-leaves the on-disk file unchanged on validation errors, while the final in-place
-write preserves its inode, mode and hard links. Adding a missing load command
-requires 16 bytes of existing header padding; ruby-macho raises
-`MachO::CodeSigningError` rather than moving segments when that space is absent.
-Only ad-hoc signing is provided: certificate identities, Developer ID signing,
-notarisation and policy assessment remain outside ruby-macho's scope. See
-[issue #262](https://github.com/Homebrew/ruby-macho/issues/262) for the original
-design discussion.
 
 ### What works?
 
--- a/README.md
+++ b/README.md
@@ -45,12 +45,49 @@ lc_vers = file[:LC_VERSION_MIN_MACOSX].first
 puts lc_vers.version_string # => "10.10.0"
 ```
 
+### Ad-hoc code signing
+
+Changing a Mach-O load command invalidates any existing code signature. This is
+especially important when Homebrew pours bottles on Apple Silicon, where native
+code must remain signed after its paths are rewritten. `MachO.codesign!` creates
+the required ad-hoc signature in Ruby instead of invoking `/usr/bin/codesign`:
+
+```ruby
+MachO.codesign!("/path/to/my/binary")
+```
+
+The implementation follows the public structures used by
+[XNU](https://github.com/apple-oss-distributions/xnu/blob/main/osfmk/kern/cs_blobs.h)
+and [ld64](https://github.com/apple-oss-distributions/ld64). For each thin
+Mach-O slice it adds or replaces `LC_CODE_SIGNATURE`, resizes `__LINKEDIT` then
+hashes the final pre-signature bytes in 4 KiB pages. It emits a SHA-256
+CodeDirectory, adding a SHA-1 alternate only when the declared deployment target
+requires legacy hash agility. Fat binaries are signed one slice at a time then
+laid out again with updated architecture offsets and sizes.
+
+When replacing a non-linker signature, the signer preserves its requirements,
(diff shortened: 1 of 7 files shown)


MikeMcQuaid:
Working as expected from https://github.com/Homebrew/brew/pull/23047

woodruffw:
(Reviewing this now.)

lib/macho/code_signing.rb
```diff
@@ -0,0 +1,576 @@
...
+      filename = File.basename(filename || "adhoc")
+      filename = File.basename(filename, File.extname(filename))
+      return filename if filename.include?(".")
+
+      uuid = macho[:LC_UUID].first
+      identity = if uuid
+        ("UUID".b + uuid.uuid.pack("C*")).unpack1("H*")
+      else
+        Digest::SHA1.hexdigest(macho.serialize.byteslice(0, macho.header.class.bytesize + macho.sizeofcmds))
+      end
```
woodruffw, line 569: Just curious, where is this identity technique documented? It seems a little strange that it's formatted as `UUID<LC_UUID>` if present and just `<UUID>` if not.

lib/macho/fat_file.rb
```diff
@@ -402,6 +412,34 @@ def repopulate_raw_machos
       end
     end
 
+    # Rebuild the fat file after internal Mach-Os change size.
+    # @return [void]
+    # @api private
+    def repopulate_resized_raw_machos
```
woodruffw, line 426: Q: why do we need this, instead of calling `repopulate_raw_machos`? That's what the other `FatFile` APIs that modify each internal `MachOFile` do.

test/test_code_signing.rb
```diff
@@ -0,0 +1,293 @@
...
+  end
+
+  private
+
+  def entitlement_blob
+    payload = "<?xml version=\"1.0\"?><plist><dict/></plist>"
+    [MachO::CodeSigning::CSMAGIC_EMBEDDED_ENTITLEMENTS, payload.bytesize + 8].pack("N2") + payload
+  end
+
+  def assert_valid_ad_hoc_signature(macho)
```
woodruffw, line 258: Nit: it'd be kind of cool if our unit tests also ran `codesign` to verify, as a kind of cross-check for correctness.

(I think we can do this only if `codesign` is available, similar to other tests that rely on external tools.)

test/test_code_signing.rb
```diff
@@ -0,0 +1,293 @@
...
+  end
+
+  def test_macos_accepts_the_signature
+    skip unless RUBY_PLATFORM.include?("darwin")
+
+    tempfile_with_data("hello", File.binread(fixture(:x86_64, "hello.bin"))) do |file|
+      MachO.codesign!(file.path)
+
+      assert system("/usr/bin/codesign", "--verify", "--strict", file.path,
+                    :out => File::NULL, :err => File::NULL)
```
MikeMcQuaid, line 137: @woodruffw Here's the test verifying with actual `/usr/bin/codesign`

lib/macho/fat_file.rb
MikeMcQuaid: Added docs for this. TL;DR: `repopulate_raw_machos` assumes slice lengths are unchanged and writes at their recorded ranges. Signing appends data, so every architecture's size and each following aligned offset must be recalculated.

lib/macho/code_signing.rb
MikeMcQuaid: @woodruffw  

- https://github.com/apple-oss-distributions/Security/blob/db15acbe6a7f257a859ad9a3bb86097bfe0679d9/OSX/libsecurity_codesigning/lib/machorep.cpp#L232-L264
- https://github.com/apple-oss-distributions/Security/blob/db15acbe6a7f257a859ad9a3bb86097bfe0679d9/OSX/libsecurity_codesigning/lib/signer.cpp#L1025-L1047uuididentity

added as comment

test/test_code_signing.rb
MikeMcQuaid: Commented below: we do have a test that verifies this.

lib/macho/code_signing.rb
woodruffw: Thanks! Funny that it's different between the two.

test/test_code_signing.rb
woodruffw: Oops, missed that. Thanks!

lib/macho/fat_file.rb
woodruffw: Oh, thanks. I forgot that our other edit operations happen within padspace, so I guess this is the first one that actually changes the real slice lengths.

README.md
```diff
@@ -45,12 +45,49 @@ lc_vers = file[:LC_VERSION_MIN_MACOSX].first
...
+
+The complete signature is built and validated before the file is written. This
+leaves the on-disk file unchanged on validation errors, while the final in-place
+write preserves its inode, mode and hard links. Adding a missing load command
+requires 16 bytes of existing header padding; ruby-macho raises
+`MachO::CodeSigningError` rather than moving segments when that space is absent.
+Only ad-hoc signing is provided: certificate identities, Developer ID signing,
+notarisation and policy assessment remain outside ruby-macho's scope. See
+[issue #262](https://github.com/Homebrew/ruby-macho/issues/262) for the original
+design discussion.
```
woodruffw: My 0.02c is we can leave this out of the README 🙂 

(It all seems factual, just a lot longer than anything else we have)

```suggestion
```

woodruffw (approved):
Nice! The implementation looks great to me.

README.md
MikeMcQuaid: @woodruffw Agree!

PR #998, opened 2026-07-10
Validate Mach-O headers before full reads

### LINKED ISSUE #23: Incremental reads for MachOFile and FatFile
Right now, `MachOFile.new` and `FatFile.new` read entire binaries into memory. This is efficient when
manipulating their contents, but is unnecessarily expensive when testing the file's sanity
(good magic, reasonable size, etc). As a result, testing large numbers of Mach-O files with exception
handling is unnecessarily slow (when using `MachOFile` or `FatFile` directly).

#22 circumvents this problem when using the generic `MachO.open` method, but the Mach-O type classes should also do this individually.

I'm assigning this to myself, but it's not particularly high on the priority list (Homebrew uses `MachO.open` only).

MikeMcQuaid:
- Avoid loading large invalid files before `MachOFile` and `FatFile` can reject malformed headers.
- Keep successful parsing on one handle so serialisation and mutation continue to use complete file data.
- Preserve compressed Mach-O handling when decompression needs the full payload.

Fixes https://github.com/Homebrew/ruby-macho/issues/23

### DIFF
--- a/lib/macho/fat_file.rb
+++ b/lib/macho/fat_file.rb
@@ -96,7 +96,12 @@ module MachO
 
       @filename = filename
       @options = opts
-      @raw_data = File.binread(@filename)
+      File.open(@filename, "rb") do |file|
+        @raw_data = file.read(Headers::FatHeader.bytesize)
+        @raw_data ||= ""
+        populate_fat_header
+        @raw_data << file.read.to_s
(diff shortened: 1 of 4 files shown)


woodruffw (approved):
Nice!

PR #1006, opened 2026-07-18
workflows/tests: replace alls-green with a conclusion job

p-linnane:
`re-actors/alls-green` is a third-party action. Replace the aggregate `check` job with a shell-based `conclusion` job that asserts the `tests` matrix result, matching `patchelf.rb` and `Homebrew/.github`.


MikeMcQuaid (approved):
Thanks! Would be good to replace as many third-party actions as possible.

PR #1028, opened 2026-09-06
Modernize repository checks and add CONTRIBUTING.md

p-linnane:
Consolidates the repository checks behind `bundle exec rake` so CI, the release workflow, and local runs all execute the same gate: RuboCop over the whole repository, the test suite, YARD with `--fail-on-warning`, and a smoke test that loads the packaged gem from a temporary directory.

Coverage minimums (94% line, 70% branch) are enforced in CI, but only for complete, unfiltered suites, so focused runs like `bundle exec ruby -Ilib -Itest test/test_macho.rb` do not fail on coverage.

Releases now verify that the tag matches the gem version and that the version is not already published before publishing.

Also fixes the `run-tests` bootstrap, which used `bundle install --path` — removed in Bundler 4 — and restricts the packaged gem to Ruby sources and documentation.

Verified locally on Ruby 3.3.12, 3.4.10, and 4.0.6.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,7 @@ permissions:
 jobs:
   release:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
 
     env:
       BUNDLE_FROZEN: "true"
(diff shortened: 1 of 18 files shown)


ZhongRuoyu (approved):

MikeMcQuaid (approved):
Thanks!

PR #1029, opened 2026-09-06
Validate ntools and nhints against cmdsize and file bounds

alebcay:
Follow-up on review findings mentioned in #1026, which found that a similar count-controlled allocation pattern exists for `LC_BUILD_VERSION#ntools` and `LC_TWOLEVEL_HINTS#nhints`.

### DIFF
--- a/lib/macho/load_commands.rb
+++ b/lib/macho/load_commands.rb
@@ -1254,15 +1254,26 @@ module MachO
       # A representation of the tool versions exposed
       # by a {BuildVersionCommand} (`LC_BUILD_VERSION`).
       class ToolEntries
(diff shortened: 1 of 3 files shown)


lib/macho/load_commands.rb
```diff
@@ -993,8 +993,12 @@ class TwolevelHintsTable
         # @param nhints [Integer] the number of two-level hints in the table
         # @api private
         def initialize(view, htoffset, nhints)
+          length = nhints * 4
```
MikeMcQuaid: ideally turn `4`, `24` etc. into variables or constants explaining the magic numbers

p-linnane (approved):
Looks good, thanks!

PR #1030, opened 2026-09-07
Validate FatArch alignment and reject truncated Fat architecture records

alebcay:
- Add parse-time validation for Fat architecture alignment exponent (align <= 15) to prevent excessive padding allocation during Fat file reconstruction and signing
- Harden `populate_fat_archs` to check record length before unpacking; truncated FatArch/FatArch64 records now raise `TruncatedFileError` instead of `NoMethodError`
- Add regression tests for truncated 32-bit and 64-bit Fat architecture records

### DIFF
--- a/lib/macho/exceptions.rb
+++ b/lib/macho/exceptions.rb
@@ -238,6 +238,15 @@ module MachO
     end
   end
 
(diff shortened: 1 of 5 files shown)


p-linnane (approved):
Awesome, thanks @alebcay!

PR #1031, opened 2026-09-07
Preserve dylib use-command flags when renaming

alebcay:
- Preserve `LC_DYLIB_USE_COMMAND` flags during dylib renames.
- Keep the original dylib use-command type instead of converting it to `LC_LOAD_DYLIB`.
- Add regression coverage for delayed-initialization and weak-link flags.


p-linnane (approved):
Nice improvement.

